PR #62, opened 2026-01-04
fix: vibe clean not working

### LINKED ISSUE #61: vibe clean does not actually remove the worktree
## Summary

The `vibe clean` command does not actually remove the worktree. It only prints the removal command to stdout instead of executing it.

## Current Behavior

When running `vibe clean` from a secondary worktree, the command outputs:

```
cd '/path/to/main/repo' && git worktree remove '/path/to/current/worktree'
```

The worktree is **not** removed. The command simply prints what should be executed.

## Expected Behavior

The `vibe clean` command should actually execute `git worktree remove` and remove the current worktree.

## Root Cause

In `src/commands/clean.ts:75`, the removal command is only logged to console:

```typescript
console.log(removeCommand);
```

There is no actual execution of the command.

## Possible Solutions

1. **Execute the command directly** - Run `git worktree remove` using `Deno.Command` before exiting
2. **Change to shell integration pattern** - If the current design is intentional (e.g., for `eval $(vibe clean)`), document this behavior clearly and consider adding a `--exec` flag for direct execution

## Additional Context

- The `post_clean` hooks are also appended to the output string but never executed
- The `pre_clean` hooks **are** executed correctly before the output

tainakanchu:
## Issue
close #61

## Summary

- Purpose:
  - Fix the `vibe clean` command to actually remove worktrees instead of just printing the command

- Background:
  - The `vibe clean` command was only outputting the removal command via `console.log()` without actually executing it
  - `post_clean` hooks were also not being executed

- Changes:
  - Modified `src/commands/clean.ts` to execute `git worktree remove` using `runGitCommand()`
  - Added `Deno.chdir(mainPath)` before worktree removal to ensure the process cwd remains valid after deletion
  - `post_clean` hooks are now properly executed via `runHooks()` from the main worktree
  - Updated tests to verify the worktree is actually removed

### Target Branch
develop

## Notes
- `pre_clean` hooks continue to work as before
- `post_clean` hooks now execute correctly from the main worktree after the secondary worktree is removed

### DIFF
--- a/src/commands/clean.ts
+++ b/src/commands/clean.ts
@@ -3,6 +3,7 @@ import {
   getRepoRoot,
   hasUncommittedChanges,
   isMainWorktree,
+  runGitCommand,
 } from "../utils/git.ts";
 import { loadVibeConfig } from "../utils/config.ts";
 import { type HookTrackerInfo, runHooks } from "../utils/hooks.ts";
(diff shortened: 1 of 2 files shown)


kexi (approved):
LGTM

Issue #408, opened 2026-04-23
feat: add vibe list command with pluggable summary generation


7tsuno:
## Summary

Add a new `vibe list` command that shows all worktrees with useful per-worktree context, including an optional user-defined "summary" column. The summary is produced by an arbitrary external command configured in `.vibe.toml`, so users can use AI (claude/gemini/ollama), simple git commands, or any custom logic.

## Problem

There is no easy way to see what each worktree is for at a glance. `git worktree list` shows path + HEAD but not the branch's intent or recent activity. When multiple worktrees accumulate, users can't tell "which one was I doing the button fix in?" without cd-ing into each.

### New command: `vibe list`

Displays all worktrees with the following columns:

| Column  | Source |
|---------|--------|
| BRANCH  | Current branch name |
| BASE    | Upstream / merge-base branch |
| AGE     | Time since last commit on branch |
| STATUS  | `clean` / `M <n>` for dirty worktrees |
| SUMMARY | Output of user-configured `[summary] command` (only if configured) |

Example:

```
$ vibe list
BRANCH                     BASE     AGE   STATUS   SUMMARY
my-feature                 develop  2h    clean    Button hover a11y fix
scratch/20260423-143052    develop  4h    M 3      Refactoring config loader
fix-bug-123                develop  1d    clean    Invalidate stale cache
```

If `[summary]` is not configured, the SUMMARY column is omitted entirely (opt-in).

### Filters & sort

Flags for narrowing and ordering the output:

| Flag | Behavior |
|------|----------|
| `--dirty` | Only worktrees with uncommitted changes |
| `--clean` | Only worktrees with no changes |
| `--base <branch>` | Only worktrees based on the given branch (e.g. `--base develop`) |
| `--recent <duration>` | Only worktrees whose last commit is within the duration (e.g. `--recent 1d`, `--recent 1w`) |
| `--stale <duration>` | Only worktrees whose last commit is older than the duration (e.g. `--stale 7d`, `--stale 2w`) |
| `--limit <n>` | Show only the top N after sort |
| `--sort <key>` | Sort order: `age` (default, newest first), `name`, `status` |
| `--reverse` | Reverse sort order |
| `--json` | Output as JSON instead of a table (see [Output format](#output-format)) |

Default ordering: most-recently-committed first (answers "what was I just doing?" naturally).

Filters combine with AND semantics (e.g., `vibe list --dirty --base develop`). `--recent` and `--stale` cannot be combined (error).

Example use cases:

```sh
[...]

PR #414, opened 2026-04-28
feat: add vibe scratch and vibe rename for unnamed worktrees

### LINKED ISSUE #409: feat: add vibe scratch and vibe rename for unnamed worktrees
## Summary

Add `vibe scratch` for creating worktrees without specifying a name (auto-generates `scratch/<timestamp>`), and `vibe rename <name>` for promoting them to proper names later. Extends `vibe list` (#408) with a `--scratch` filter and adjusts `vibe jump` to skip scratches in fuzzy matching.

## Problem

When you want to "just try something" in a new worktree, coming up with a branch name is friction. Users end up with:

- Throwaway names like `test`, `tmp`, `asdf` polluting the branch list
- Abandoning the idea before even starting, because naming feels heavyweight
- Committing to a name too early, then wanting to rename once the work takes shape

The core pain: **naming should happen when you have something worth naming**, not upfront.

### `vibe scratch`

Create a worktree with an auto-generated name; no argument required.

```
$ vibe scratch
Created worktree: scratch/20260423-143052
```

Details:
- Branch name: `scratch/<YYYYMMDD-HHMMSS>` (local-time timestamp, predictable and sortable)
- Behavior otherwise matches `vibe start`: same base branch resolution, same `post_start` hooks, same CoW copy, same trust flow
- Output format matches `vibe start` — no new concepts introduced

Why a separate command instead of `vibe start` with no args:
- Explicit intent: "this is a scratch, I know what I'm doing"
- Avoids accidents when users mistype `vibe start`
- Leaves room for scratch-specific flags later

### `vibe rename <name>`

Rename the current worktree's branch and directory to the given name. Primary use: promote a scratch to a proper name once the work crystallizes.

```
$ vibe rename my-feature
Renamed scratch/20260423-143052 → my-feature
Directory: /Users/.../worktrees/scratch-20260423-143052 → /Users/.../worktrees/my-feature
```

Details:
- Runs `git branch -m <old> <new>` + moves the worktree directory
- Updates any internal vibe state (trust entries, cached summaries keyed by name)
- Usable on any worktree, not just scratches — scratch-promotion is just the primary use case
- Target name must not already exist as a branch or directory

**Pushed branches are not supported.** If the current branch has an upstream, `vibe rename` aborts and prints the exact git commands for manual handling:

```
$ vibe rename my-feature
error: branch 'scratch/20260423-143052' is pushed to 'origin'
vibe rename does not handle pushed branches to avoid remote divergence.

To rename manually:
  git branch -m scratch/20260423-143052 my-feature
  git push origin -u my-feature
  git push origin --delete scratch/20260423-143052
  # then move the worktree directory yourself
```

This keeps vibe's operation simple and predictable; users who need to rename remote branches do it consciously with git directly.

### `vibe list --scratch` (extends #408)

Add a filter to the list command introduced in #408:

| Flag | Behavior |
|------|----------|
| `--scratch` | Only worktrees with a `scratch/` branch prefix |

Scratches are included in the default `vibe list` output (they are your worktrees, after all). `--scratch` is for when you want to focus on them, e.g. `vibe list --scratch --stale 7d` to find stale scratches for cleanup.

### `vibe jump` behavior for scratches

`vibe jump` with fuzzy matching **skips scratches by default** to keep the candidate list clean. To jump to a scratch:

- Type the full prefix: `vibe jump scratch/20260423-143052` (explicit match always works)
- Or use `vibe list --scratch` to find the name, then `vibe jump <name>`

Rationale: scratches accumulate and would pollute fuzzy-match results. Named branches are the primary jump targets; scratches are secondary by design.

### Why `scratch/` prefix

- `scratch/` is visually distinct in `git branch` output
- The `/` makes it a "namespace" that tools (including git itself) can treat as a group
- Shorter than `experimental/`, clearer than `tmp/`

### Why timestamps instead of Docker-style names

Timestamps (`scratch/20260423-143052`) are:
- Sortable (newest first in `vibe list`)
- Collision-free by construction
- Easier to associate with "when I did that thing"

Docker-style names (`happy-fox`) are memorable but offer no temporal signal and occasionally produce awkward combinations.

## Out of Scope

- "Resume most recent scratch" magic (`vibe scratch` with no args when inside another worktree). Users can `vibe list --scratch` + `vibe jump`
- Auto-cleanup of old scratches. Users run `vibe list --scratch --stale 30d` then `vibe clean` manually
- Renaming pushed branches (handled manually via git, as documented above)
- Anonymous/detached HEAD worktrees (the "separated" design alternative we explored). Integrated approach chosen for simplicity

## Open Questions

1. Should `vibe rename` support renaming *another* worktree (not just the current one)? E.g., `vibe rename scratch/foo my-feature`. Suggested: **current only** at first; single-arg form is simpler and handles the main use case
2. Should scratch creation print a hint like `(run 'vibe rename <name>' to promote)` after success? Probably yes, low cost for discoverability
3. Should `vibe clean` treat scratches specially? Suggested: **no special treatment**; existing `vibe clean` already handles them uniformly

## Acceptance Criteria

- [ ] `vibe scratch` creates a worktree named `scratch/<timestamp>` and cd-auto works
- [ ] `vibe scratch` runs existing `pre_start` / `post_start` hooks
- [ ] `vibe rename <name>` renames both the git branch and the worktree directory
- [ ] `vibe rename` aborts with git-command guidance if the branch has an upstream (pushed)
- [ ] `vibe rename` aborts if the target name already exists
- [ ] `vibe list --scratch` shows only scratches
- [ ] `vibe jump` fuzzy-match does not surface scratches as candidates
- [ ] `vibe jump scratch/<full-name>` still works via explicit match
- [ ] Docs (en + ja) updated for both commands and the new `--scratch` filter

## Dependencies

Depends on #408 (for `vibe list` and the `--scratch` filter).

7tsuno:
## Summary

Implements [#409](https://github.com/kexi/vibe/issues/409) (excluding the `vibe list --scratch` filter, which depends on #408):

- **`vibe scratch`** creates a worktree with an auto-generated `scratch/<YYYYMMDD-HHMMSS>` branch name, delegating to `startCommand` once a unique name is resolved. Same hooks/copy/CoW flow as `vibe start`. Collisions in the same second append a `-2`/`-3`/... suffix (capped at 100).
- **`vibe rename <new-name>`** renames the current worktree's branch and moves its directory in one step. Push-tracked branches are rejected with the exact git commands needed to do it manually; the main worktree is also rejected. `git worktree move` runs first so the metadata gets updated by git itself, then `git branch -m`; if the branch rename fails after the move succeeds, the move is rolled back. MRU is updated in place so jump ordering is preserved.
- **`vibe jump`** filters worktrees on a `scratch/`-prefixed branch out of word-boundary, substring, and fuzzy match candidates. Exact matches still surface them, and queries that themselves start with `scratch/` (case-insensitive) opt back into the full pool.

`vibe list --scratch` is intentionally out of scope and tracked under #408.

### Notable design choices

- `scratch.ts` exports `SCRATCH_PREFIX` so `jump.ts` doesn't have to redefine the constant.
- `scratch` collision-loop fetches all worktrees and existing scratch branches once via `Promise.all` (`getWorktreeList` + `git for-each-ref refs/heads/scratch/`), then checks against an in-memory `Set` — the loop itself does not spawn extra git subprocesses.
- `rename` derives the new path from `getMainWorktreePath`, not the cwd-based `getRepoRoot`, so renaming from inside a scratch produces `<repo>-<branch>` (e.g. `vibe-feat-hoge`) rather than `<scratch-name>-<branch>`.
- After `git worktree move`, `runtime.control.chdir(newPath)` is called so subsequent git subprocesses don't fail with `ENOENT, posix_spawn 'git'` (the inherited cwd no longer exists).
- Branch names keep their slashes (`feat/hoge`); only the on-disk directory is sanitized (`vibe-feat-hoge`), matching what `vibe start` does.
- MRU `updateMruBranch` preserves the existing entry's timestamp so MRU ordering survives a rename.

## Commits (logical units)

1. `feat: add vibe scratch command for auto-named worktrees`
2. `feat: add vibe rename command for branch and directory rename`
3. `feat: exclude scratch worktrees from vibe jump non-exact matching`
4. `feat: wire vibe scratch and rename commands into CLI`
5. `docs: list vibe scratch and rename in command index and READMEs`
6. `fix: keep vibe rename runnable from secondary worktrees`
7. `fix: keep slashes in vibe rename branch names`

## Test plan

- [x] `pnpm run check:all` green (lint, prettier, tsc, vitest, astro check, video typecheck)
- [x] 451 unit tests passing (timestamp / mru.updateMruBranch / worktree rename services / scratchCommand / renameCommand / jumpCommand scratch exclusion)
- [x] Manual E2E: `vibe scratch` -> `scratch/<ts>` worktree created
- [x] Manual E2E: same-second second invocation produces `-2` suffix
- [x] Manual E2E: `vibe rename feat/hoge` from a scratch -> branch `feat/hoge`, dir `vibe-feat-hoge`
- [x] Manual E2E: `vibe rename` from main worktree rejected
- [x] Manual E2E: `vibe rename` rolls back cleanly when `git branch -m` fails
- [x] Manual E2E: `vibe jump <fuzzy>` excludes scratch candidates; `vibe jump scratch/...` opts in
- [x] docs (en/ja) build via astro check

Closes #409 (excluding `vibe list --scratch`, which depends on #408)


kexi (approved):
LGTM!

PR #504, opened 2026-06-07
chore: probe required nix check

kexi:
Temporary draft PR to verify whether the required nix-build check blocks when the Nix workflow is skipped by path filters. This PR contains only an empty commit and should be closed after observation.


kexi:
Closing this probe PR now that #505 verifies the required Nix check behavior.

PR #582, opened 2026-08-02
fix: reproduce elected-license notices and attach license docs to releases

kexi:
## Why

An audit of the v3.0.0 distribution channels found two compliance gaps against the MIT license's "shall be included in all copies or substantial portions" condition:

1. **THIRD-PARTY-LICENSES.md** only reproduced license texts for the four crates whose SPDX expression is a top-level `AND`. Crates where vibe elects MIT out of `MIT OR Apache-2.0` (and pure-MIT crates) were listed as bare SPDX identifiers — but electing MIT is precisely what makes MIT's notice condition bind, so the copyright and permission notices of ~160 statically linked crates were missing from every shipped copy.
2. **GitHub Release assets** were bare binaries with no license text at all. Homebrew and the Nix prebuilt package consume those same assets, so no binary channel outside npm and the .deb carried any notices.

npm (shim + platform packages) and the .deb were already compliant and are unchanged.

## What

- `scripts/generate-third-party-licenses.ts`: parse each crate's SPDX expression and elect one arm — a notice-free license (Unlicense / CC0-1.0 / MIT-0 / 0BSD) when available, otherwise MIT-first among attribution-bearing options. The elected license's text is reproduced verbatim from the crate's shipped license file in a new **Appendix B** (161 entries; 12 crates ship no license file and fall back to the canonical MIT text derived from their `Cargo.toml` declaration, with a note). Top-level-`AND` crates keep the existing full reproduction as Appendix A. All existing safety machinery (symlink containment, UTF-8/control-character validation, 1 MB cap, adaptive fences) is reused on the new path.
- `THIRD-PARTY-LICENSES.md`: regenerated, 1,303 → 6,325 lines.
- `release.yml` / `beta-release.yml`: attach `LICENSE` and `THIRD-PARTY-LICENSES.md` (from the release commit's checkout) as release assets, and add a named verification step running the new `scripts/verify-release-assets.ts`, which fails the run if either document is missing, empty, or not fully uploaded — the compliance gap is now CI-enforced instead of manually audited.
- Tests: unit tests for the election logic and the verifier, plus a cross-check invariant that every crate in the table whose elected license bears an attribution obligation has an Appendix B section.
- `CLAUDE.md`: sync the `just check` description with what `check:all` actually runs.

## Verification

- `just check` (`pnpm run check:all`) passes end to end, including `check:licenses` freshness and the 224 npm tests.
- Regeneration is idempotent (byte-identical on a second run).
- The release-asset verifier runs in the `release.yml` dry-run path too, so the next dry run exercises it before any real release.

https://claude.ai/code/session_01AFQdQUb1tdTv7G6YvmWu93

## Summary by CodeRabbit

- **Documentation**
  - Clarified third-party license obligations, notices, and attribution details.
  - Added elected license texts and organized notices into separate appendices.
  - Updated release guidance for license verification.

- **Release Improvements**
  - Release packages now include project and third-party license documentation.
  - Releases are verified before publication to ensure required license files are present, complete, and correctly uploaded.

- **Quality Improvements**
  - Expanded coverage for license generation, attribution, release assets, and incomplete or malformed release data.
  - Improved handling of license notices and attribution text.


kexi:
Triage of the 5 CodeRabbit findings (also cross-checked by a second reviewer):

- **Appendix B preamble** — fixed in 372232e: the "statically linked" claim is softened and Appendix B now carries the same conservative-graph / platform-gated-crates disclaimer as Appendix A.
- **Notice-file matcher (bidirectional prefix)** — already fixed in a95f638: one-directional token-sequence prefix matching with a known-license guard; a regression test covers elected `MIT` vs `LICENSE-MIT-0`.
- **Missing size/state metadata in verify-release-assets** — fixed in 372232e: absent `size` or `state` on a required asset is now reported as its own problem (the real producer, `gh release view --json assets`, always emits both; treating "cannot tell" as a pass is backwards for a release gate). Negative sizes are caught too.
- **` ```text ` fence language** — declined: fences are length-adaptive to survive backtick runs inside license texts, GitHub renders bare fences as plain text already, and the info string would regenerate a 6,000-line document for a purely cosmetic change.
- **`--help` / `--version` for verify-release-assets** — declined: it is an internal CI gate invoked only from the two release workflows, not a user-facing CLI; its stdin/file contract is documented in the header.

PR #584, opened 2026-08-02
feat: protect the default branch from clean and rename

### LINKED ISSUE #578: feat: protect the default branch from clean and rename
## Motivation

Nothing currently stops `vibe clean --delete-branch` or `vibe rename` from operating on the repository's default branch (e.g. `main` or `develop`). Deleting or renaming the default branch is almost never intended and can be disruptive to the whole team, especially since `clean --delete-branch` removes the branch after the worktree.

## Proposal

- Resolve the default branch dynamically instead of hardcoding names:
  1. `git symbolic-ref refs/remotes/origin/HEAD --short` (strip the `origin/` prefix)
  2. Fallback: `git config --get init.defaultBranch`
  3. Fallback: `master`
- `vibe clean` on the default branch's worktree: remove the worktree but keep the branch, with a message explaining why (even when `--delete-branch` is passed).
- `vibe rename` on the default branch: refuse with an error.
- Add an escape hatch flag (e.g. `--allow-default-branch`) for the rare cases where the operation is intentional.

## Notes

- The resolution logic fits naturally next to the existing git helpers in `vibe-core`.
- Behavior differences (soft-skip for branch deletion vs hard error for rename) should be covered by unit tests and E2E tests.

kexi:
## Summary

Protects the repository's default branch from `vibe clean --delete-branch` and `vibe rename`, as proposed in #578.

- **Dynamic resolution** — `get_default_branch` in `vibe-core/src/git.rs` resolves the name from `git symbolic-ref refs/remotes/origin/HEAD --short` (stripping the `origin/` prefix), falling back to `git config --get init.defaultBranch`, then to `master`. Every step is best-effort: a guard that failed hard would break `clean` and `rename` in repositories where git has no opinion about a default branch. Empty answers (a bare `origin/`, a blank config value) are skipped rather than adopted, so the guard can never match every branch.
- **`vibe clean` soft-skips** — the worktree is still removed and the command still exits 0; only the branch deletion is skipped, with a message explaining why. The branch is incidental to `clean`, so a successful cleanup should not be turned into a failure.
- **`vibe rename` hard-errors** — the branch change *is* the product of `rename`, so there is nothing useful left to do. The guard is placed before the pushed-branch guard so the message names the real reason, including for a default branch in a repo with no remote.
- **`--allow-default-branch`** on both commands is the escape hatch. It is wired through clap, the completion spec, and the fish/zsh completion snapshots.

## Tests

- `git.rs`: 5 unit tests covering each resolution tier, `origin/` prefix stripping, branch names containing slashes, and empty-answer fallthrough.
- `clean_tests.rs`: 3 unit tests — default branch soft-skipped (worktree still removed, branch survives), `--allow-default-branch` deletes it, non-default branches unaffected.
- `rename.rs`: 4 unit tests — refused without mutation, `--allow-default-branch` reaches the success path, precedence over the pushed-branch guard, non-default branches unaffected.
- `packages/e2e/default-branch.test.ts`: 6 new E2E tests driving the real binary through node-pty, including one that verifies `origin/HEAD` wins over `init.defaultBranch`.

## Checks run

All of `pnpm run check:all`'s constituent checks passed locally:

- `fmt:check`, `lint`, `check:i18n`, `check:licenses`, `test:npm`, `check:docs` — via `pnpm run <script>`
- `check:rust` — `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` (605 vibe-core tests + the binary suites, all green)
- `test:e2e` — the full `packages/e2e` suite, 51 passed / 1 skipped, including the 6 new tests

The cargo and e2e steps were run with `CARGO_TARGET_DIR` pointed at an existing warm target directory and `VIBE_BINARY_PATH` at the resulting debug binary, because the sandbox host had insufficient free disk for a second full target tree. The commands and their inputs are otherwise identical to what `check:all` runs.

## Follow-up

No user-facing docs were touched to keep this PR code-focused. `packages/docs/src/content/docs/commands/{clean,rename}.mdx` (and the `ja/` + `zh/` mirrors) should document `--allow-default-branch` in a follow-up.

Closes #578

### DIFF
--- a/packages/e2e/default-branch.test.ts
+++ b/packages/e2e/default-branch.test.ts
@@ -1,0 +1,197 @@
+import { execFileSync } from "child_process";
+import { existsSync } from "fs";
+import { basename, dirname } from "path";
+import { afterEach, describe, expect, test } from "vitest";
+import { getVibePath, VibeCommandRunner } from "./helpers/pty.js";
+import { setupTestGitRepo } from "./helpers/git-setup.js";
+import {
+  assertExitCode,
+  assertNonZeroExitCode,
+  assertOutputContains,
+} from "./helpers/assertions.js";
+
+/**
+ * Guards `vibe clean --delete-branch` and `vibe rename` against operating on the
+ * repository's default branch (issue #578).
+ *
+ * `setupTestGitRepo` builds a repo with no remote, so the default branch is
+ * resolved from `init.defaultBranch`; each test declares it explicitly rather
+ * than relying on whatever the host's global git config happens to say.
+ */
+
+function git(args: string[], cwd: string): string {
+  return execFileSync("git", args, { cwd, stdio: "pipe", encoding: "utf-8" });
+}
+
+function branchExists(repoPath: string, branchName: string): boolean {
+  try {
+    git(["show-ref", "--verify", "--quiet", `refs/heads/${branchName}`], repoPath);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+/**
+ * Add a worktree checked out on a NEW branch and return its path.
+ */
+function addWorktree(repoPath: string, branch: string, suffix: string): string {
+  const worktreePath = `${dirname(repoPath)}/${basename(repoPath)}-${suffix}`;
+  git(["worktree", "add", "-b", branch, worktreePath], repoPath);
+  return worktreePath;
+}
+
+describe("default branch protection", () => {
+  let cleanup: (() => Promise<void>) | null = null;
+
+  afterEach(async () => {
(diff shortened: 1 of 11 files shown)


kexi:
## Codex review (gpt-5.6-sol)

No findings — LGTM.

Codex reviewed default-branch protection logic, error handling/exit-code contracts, eval/output safety, git argument handling, cross-platform behavior, and test coverage, and found no P1/P2/P3 issues.

*(Automated review via Codex CLI, coordinated by Claude.)*

PR #604, opened 2026-08-15
fix: avoid panic on non-ASCII hash strings in verify

### LINKED ISSUE #596: fix: verify panics on multibyte hash strings in settings.json
## Summary

`vibe verify` panics when an `AllowEntry.hashes` value in `settings.json` contains multibyte characters.

**Location**: `rust/crates/vibe-core/src/commands/verify.rs:178`

```rust
let prefix = &hash[..hash.len().min(16)];
```

## Failure scenario

1. `~/.config/vibe/settings.json` contains an allow entry whose `hashes` array has a non-hex multibyte string (e.g. `"日本語のハッシュ値です"`, 33 bytes). This can happen via hand-editing, corruption, or a broken migration — `AllowEntry.hashes: Vec<String>` is deserialized with no hex validation.
2. Run `vibe verify` in the target repository, reaching the Hash History loop in `display_file_status`.
3. `hash.len()` is the **byte** length (33), `min(16)` slices at byte 16, which falls inside a multibyte char boundary:

```
panicked at crates/vibe-core/src/commands/verify.rs:178:27:
byte index 16 is not a char boundary; it is inside 'ッ' (bytes 15..18) of `日本語のハッシュ値です`
```

Since the release profile uses `panic = "abort"`, this bypasses `VibeError` formatting and exit-code handling entirely.

## Suggested fix

Use a char-boundary-safe truncation, e.g. `hash.chars().take(16).collect::<String>()`. Optionally validate `hashes` entries as hex at deserialization time.

Note: the similar slice at `doctor.rs:729` operates on an ASCII `r#"` delimiter length and is safe.

Severity: high / reproduced.

kexi:
## Summary

`vibe verify` aborted instead of reporting status when an `AllowEntry.hashes` value in `~/.config/vibe/settings.json` contained multibyte characters.

`AllowEntry.hashes` is deserialized as a plain `Vec<String>` with no hex validation, so a hand-edited, corrupted, or badly migrated entry can hold arbitrary text. The Hash History loop in `display_file_status` truncated each entry with a **byte** slice:

```rust
let prefix = &hash[..hash.len().min(16)];
```

For e.g. `"日本語のハッシュ値です"` (11 chars / 33 bytes) byte index 16 falls inside `'ッ'` and the slice panics. Since the release profile uses `panic = "abort"`, this bypassed `VibeError` formatting and exit-code handling entirely.

The fix replaces the byte slice with a char-boundary-safe `hash_prefix()` helper (`hash.chars().take(16).collect()`). This is byte-identical for real ASCII hex hashes, so the `verify` output contract is unchanged; a multibyte entry now renders its first 16 characters and `verify` continues to the normal `❌ HASH MISMATCH` report.

Audit: `verify.rs` had no other byte slice of a settings-derived string (the only other occurrence was in a test assertion, now routed through the same helper). As noted in the issue, `doctor.rs:729` slices at an ASCII `r#"` delimiter length and is safe.

Out of scope (follow-up): validating `hashes` entries as lowercase hex at deserialization time in `settings.rs` would reject such data at the source rather than only rendering it safely. That is a behaviour change to settings loading (it would have to decide between rejecting the whole settings file and dropping the bad entry), so it is deliberately left out of this crash-only fix.

Fixes #596

## Test Plan

Added unit tests in `rust/crates/vibe-core/src/commands/verify.rs`:

- `multibyte_hash_history_entry_is_truncated_by_chars_without_panicking` — drives a multibyte `hashes` entry through the full `verify_command` / `display_file_status` path (the exact repro from the issue) and asserts it renders and reports `HASH MISMATCH` rather than panicking.
- `long_multibyte_hash_history_entry_is_cut_at_char_16` — a >16-char multibyte entry is cut at the 16th character boundary, not at byte 16.
- `hash_prefix_truncates_by_chars_and_preserves_ascii_rendering` — ASCII hashes keep their previous rendering; short/empty/exactly-16-char inputs behave.

Ran locally, all passing:

```
cd rust && cargo fmt --all --check
cargo clippy --workspace --all-targets -- -D warnings
cargo test -p vibe-core     # 729 passed
```

Full CI validates the rest of the check suite on this PR.

## Checklist

- [x] Tests added/updated
- [ ] `pnpm run check:all` passes (targeted Rust checks run locally; full suite via CI)
- [x] Docs updated (if needed) — no user-facing behaviour change to document


kexi:
## Review convergence loop: converged in 1 round

**Reviewer**: OpenAI Codex CLI (`gpt-5.6-sol`, reasoning effort `high`, read-only sandbox), reviewing `git diff origin/develop...HEAD`. Triage and any fixes by Claude.

- **Rounds run**: 1
- **Findings fixed**: none — Codex returned `[]` (no findings) on the first round, so no commits were made.
- **Findings disputed**: none.
- **Convergence**: converged (clean first round).

**Local verification on the branch head** (`20fca31`):

- `cargo fmt --all --check` — pass
- `cargo clippy --workspace --all-targets -- -D warnings` — pass
- `cargo test -p vibe-core` — 729 passed, 0 failed

kexi:
@coderabbitai review

kexi:
@coderabbitai review

PR #605, opened 2026-08-15
fix: distinguish failure and abort glyphs in progress tree

### LINKED ISSUE #600: fix: progress tree renders success, failure, and abort with the same glyph (☒)
## Summary

`complete_task` (success), `fail_task` (failure), and `finish` (aborted/incomplete) all render `☒`, so users cannot tell success from failure on screen.

**Locations**: `rust/crates/vibe-core/src/progress.rs:154`, `:163`, `:175`

```
IndicatifTracker::complete_task → "{prefix}☒ {label}"        ← success
IndicatifTracker::fail_task     → "{prefix}☒ {label} ({msg})" ← failure
IndicatifTracker::finish        → "{prefix}☒ {label}"        ← aborted
```

## Failure scenario (reproduced via pty)

`vibe start <branch> --reuse` shows `┗ ☒ assets` when the copy **succeeds** (exit 0, `cd` emitted) — and the exact same rendering when it actually **fails** (read-only destination, `Permission denied`, exit 1, empty stdout).

## Background

This is a port regression: the removed TypeScript `TreeFormatter` (`packages/core/src/utils/progress.ts`) distinguished them:

```ts
static readonly COMPLETED = "☒";   // success
static readonly FAILED    = "✗";   // failure (red)
```

No test asserts the rendered strings (`RecordingTracker` records event kinds only), which is why this slipped through CI.

## Suggested fix

Restore `✗` (red) for `fail_task`, consider a distinct glyph for `finish`-with-incomplete-tasks, and add a rendering assertion test.

kexi:
## Summary

`IndicatifTracker` rendered success (`complete_task`), failure (`fail_task`) and the
aborted-run sweep (`finish`) with the same `☒` glyph, so a failed copy or hook was
visually indistinguishable from a successful one — `vibe start <branch> --reuse` printed
`┗ ☒ assets` whether the copy succeeded or died with `Permission denied`. This was a
port regression: the removed TS `TreeFormatter` used `COMPLETED = "☒"` and a red
`FAILED = "✗"`.

Changes:

- The final line of a progress node is now produced by the pure
  `render_line(TaskOutcome, prefix, label, color)` in
  `rust/crates/vibe-core/src/progress.rs` (the failure reason travels inside
  `TaskOutcome::Failed { error }`, so a failure line cannot be rendered without one):
  - `TaskOutcome::Completed` → `<prefix>☒ <label>` (never colored, unchanged rendering)
  - `TaskOutcome::Failed` → `<prefix>✗ <label> (failed: <err>)` in RED
  - `TaskOutcome::Abandoned` → `<prefix>⊘ <label>` in DIM
- `finish` no longer fake-completes tasks that were still pending. Which outcome each
  still-open node gets is decided by the pure `closing_outcomes`, which knows the node's
  kind: a task still pending is abandoned (`⊘`, dim), while a **phase** header closes as
  done unless a task under it was still pending — a phase is never completed by a caller,
  so treating it like a task would paint every header of a fully successful run as
  abandoned.
- The abandoned marker is `⊘`, not `☐`: the README/docs legend already spends `☐` on
  "queued, will still run", so reusing it would make "gave up" and "not started yet" the
  same glyph. All three locales of the README and the hooks docs now carry an explicit
  marker legend.
- Color goes through the existing `ansi::colorize` rather than hardcoded escapes.
  `IndicatifTracker::with_color(..)` takes the caller's already-resolved
  `ansi::is_color_enabled(io)` value, so `NO_COLOR` / `FORCE_COLOR` are honoured for
  the progress tree too; both wiring sites (`vibe/src/commands/mod.rs::pick_tracker`
  and `vibe-core/src/commands/mod.rs::RealStart`) pass it. `IndicatifTracker::new()` /
  `Default` are gone so the color decision cannot be silently guessed.
- **Security (audit follow-up):** `render_line` now neutralizes both `label` and the
  failure reason with `output::sanitize_for_display`. The line wraps caller text in ANSI
  codes and draws it onto a live, redrawable display, but most call sites passed raw
  repository-controlled text — the hook command strings from `.vibe.toml`
  (`commands/start.rs`, `commands/clean.rs`) and the copy pattern (`copy/symlink.rs`).
  Only `copy_runner` sanitized, so the same hook string was neutralized on one stderr
  path and not on another; a trusted-but-hostile hook name embedding `\x1b[2K\x1b[A` or
  a bidi override could rewrite the progress tree and forge lines the run never produced
  (display spoofing only — the stdout/eval contract is unaffected). Sanitizing inside
  `render_line` rather than at each caller means a future `add_task`/`fail_task` caller
  cannot reopen the hole by forgetting. The existing `copy_runner` sanitization stays,
  because those same strings also reach plain stderr through `warn_log`/`log_dry_run`,
  which never pass through `render_line`; `sanitize_for_display` maps unsafe characters
  to U+FFFD and is therefore idempotent, so the double pass is a no-op.
- Updated the progress section of `.claude/agents/vibe-develop-expert.md`, which
  documented the old identical-glyph behaviour.

Success rendering is byte-identical to before for ordinary labels, so the `☒` examples in
README / docs still hold.

## Test Plan

The issue slipped through CI because `RecordingTracker` records only event kinds, never
the rendered string. Extracting `render_line` as a pure function gives a test seam for
the actual glyphs; added unit tests in `rust/crates/vibe-core/src/progress.rs`:

- `success_failure_and_abandon_render_distinct_glyphs` — asserts the three exact
  rendered strings and that they are pairwise different (the regression guard).
- `abandoned_does_not_reuse_the_documented_pending_glyph` — `⊘` never collides with `☐`.
- `failure_is_red_and_abandon_is_dim_when_color_is_enabled` — asserts the RED / DIM
  wrapping with reset.
- `success_is_never_colored` — success stays plain even with color on.
- `color_is_suppressed_when_disabled` — no ANSI escapes in any outcome when color is off.
- `prefix_and_tree_shape_are_preserved` — the `┗ ` tree prefix survives.
- `finish_*` — drive the tracker's real `add_phase`/`add_task`/`complete_task`/
  `fail_task` sequence and assert what `finish()` paints on every line: a successful
  phase stays done, a failed task keeps its own `✗` line while its phase still closes,
  a pending task and its phase are abandoned, a task-less phase completes, and pending
  tasks are scoped to their own phase.
- `labels_cannot_carry_control_sequences_for_any_outcome` — a label containing
  `\x1b[2K\x1b[A` and U+202E renders with those bytes replaced by U+FFFD for
  completed / failed / abandoned, with color on and off. Guarantees a progress line
  cannot carry a raw control sequence out of a label.
- `failure_reason_cannot_carry_control_sequences` — the same for the `(failed: …)` text.
- `sanitize_is_idempotent` — sanitizing twice equals sanitizing once, which is what lets
  `copy_runner` keep its own call-site sanitization alongside the central one.

Ran locally and passing:

```
cd rust && cargo fmt --all --check \
  && cargo clippy --workspace --all-targets -- -D warnings \
  && cargo test --workspace
```

Full `check:all` CI validation runs on this PR.

## Checklist

- [x] Tests added/updated
- [ ] `pnpm run check:all` passes (targeted Rust checks run locally; full CI validates on this PR)

Fixes #600

## Summary by CodeRabbit

* **New Features**
  * Progress displays now distinguish running, completed, failed, and abandoned tasks with clear markers and colors.
  * Spinners stop cleanly when tasks or phases finish, including failed and interrupted operations.
  * Progress labels and error messages are safely formatted for terminal display.

* **Documentation**
  * Updated English, Japanese, and Chinese guides with revised hook output examples and a legend explaining each progress marker.


kexi:
## Review convergence loop — 3 rounds

Reviewer: **Codex CLI** (`gpt-5.6-sol`, reasoning effort high), read-only, reviewing `git diff origin/develop...HEAD`. Fixes by **Claude**. Every finding was verified against the code before being applied.

### Findings fixed (4)

- **`indicatif` stamped a spinner glyph on every closed line** (`progress.rs`). `{spinner}` on a *finished* bar expands to the last `tick_strings` entry, not to nothing, so outcome lines were drawn as `⠏ ┗ ✗ assets (failed: …)` — a running glyph in front of the marker that says the task stopped. Clearing the prefix could not help, since the placeholder lives in the template. `close_bar` now swaps the node to a `{msg}`-only `closed_style()` (with blanked tick strings) before painting. Confirmed against indicatif 0.17.11 `style.rs:167-182`.
- **The `finish_*` tests could not have caught it.** The helper re-derived its expectations by calling `closing_outcomes` + `render_line` again, so it agreed with the implementation by construction. It now drives the real `tracker.finish()` and reads back each bar's actual message; already-closed nodes are reported as `<kept:…>` carrying their real line, which additionally pins that `finish()` does not repaint over a `✗` that `fail_task` produced.
- **The marker legend documented a glyph the tool cannot emit.** The `☐` "queued" row had no code path behind it — a node carries the spinner style from the moment it is added, and `start_task` only begins animating it. Removed from all three READMEs and all three `hooks.mdx`; `⠋` now covers pending and running.
- **`.claude/agents/vibe-develop-expert.md` cited the removed `☐` row** to justify `⊘`, and described the spinner template as the only style a node wears. Both now match the code, plus the new tests.

### Findings disputed

None. All four findings were reproducible.

### Regression guards added

`closed_nodes_stop_rendering_the_spinner` asserts `get_final_tick_str() == ""` for every node closed by `complete_task`, `fail_task` or `finish`; `the_spinning_style_would_stamp_a_glyph_on_a_finished_bar` pins that the *running* style's final tick really is `⠏`, so the first test guards a live hazard rather than a hypothetical one. Verified by reverting the fix: the guard fails with `left: "⠏", right: ""`.

### Convergence

Round 3 returned `[]` — no findings, nothing outstanding.

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (744 + 27 + 14 + 13 + 6 + 6 + 1 passing), `check:i18n` and `fmt:check` all pass.

kexi:
@coderabbitai review

PR #606, opened 2026-08-15
fix: apply *_prepend/*_append in single-file config loads

### LINKED ISSUE #599: fix: *_prepend/*_append fields silently ignored when only .vibe.toml exists
## Summary

The `_prepend`/`_append` fields are only expanded by `merge_configs`, but `load_vibe_config` only calls `merge_configs` when **both** `.vibe.toml` and `.vibe.local.toml` exist. With `.vibe.toml` alone (the most common setup), all 12 `*_prepend`/`*_append` fields are silently dropped.

**Location**: `rust/crates/vibe-core/src/config_loader.rs:45-50` (root cause: the single-file path bypasses `config.rs::merge_configs`)

## Failure scenario (reproduced with a test)

1. A repo has only `.vibe.toml` (no `.vibe.local.toml`):
   ```toml
   [copy]
   files = [".env"]
   files_append = [".env.local"]
   ```
2. `load_vibe_config` hits the `(Some(base), None) => Some(base)` arm and returns the parsed config as-is.
3. Consumers only read `config.copy.files`; `_prepend`/`_append` fields are referenced nowhere outside `merge_configs` and tests.
4. `.env.local` is not copied — no warning, no `deny_unknown_fields` error (the field is valid per schema). Adding a `.vibe.local.toml` suddenly changes behavior, which makes this very confusing to reproduce as a user.

## Design note

Simply routing the single-file path through `merge_configs(&default, &cfg)` is not enough: `merge_array_field` prefers the override side, so `files` + `files_append` coexisting in the same file would drop the `_append` values. The semantics of base + append **within one file** need an explicit decision first.

kexi:
## Summary

Fixes #599

Every `*_prepend` / `*_append` field is accepted by the `.vibe.toml` schema, but only `merge_configs` expanded them, and `load_vibe_config` only reached `merge_configs` when **both** `.vibe.toml` and `.vibe.local.toml` existed. In the common single-file setup those fields were parsed and then silently discarded — no warning, no `deny_unknown_fields` error — and adding a `.vibe.local.toml` suddenly changed behaviour.

The fix makes within-file extension resolution a first-class step shared by every load path:

- `config.rs::merge_array_field` now wraps the effective base `override.or(base)` with the same source's prepend/append, instead of returning an explicit override outright. Result: `prepend ++ (override | base) ++ append`.
- New pure `config.rs::normalize_config(cfg) = merge_configs(&VibeConfig::default(), cfg)` resolves one file's own triplets, so single-file and two-file loading share **one** implementation of the array algebra.
- `config_loader.rs::load_vibe_config` routes every arm through it: single-file loads become `normalize_config(&file)`, the two-file load becomes `merge_configs(&normalize_config(&base), &local)`.

The local file is deliberately **not** pre-normalized in the two-file arm: an extension-only local (e.g. only `files_append`) must reach `merge_array_field` with its field slot still `None` so it *wraps* the base's effective array instead of overriding it. That is the documented cross-file merge table and is pinned by the untouched `trusted_base_and_local_are_merged` test.

### Intentional behaviour changes (each turns a silent drop into the documented effect)

1. Single-file `*_prepend` / `*_append` now apply (the reported bug).
2. Base-file `*_prepend` / `*_append` now fold into the base's effective array before the local merge.
3. A file setting both `field` and `field_prepend` / `field_append` now gets `prepend ++ field ++ append` instead of `field` alone.
4. Edge case now documented: `files = []` (the disable idiom) together with `files_append = ["x"]` in the same source yields `["x"]`, not `[]`.

Non-array fields (`concurrency`, `untracked`, `modified`, `worktree.path_script`, `clean.delete_branch`, `submodules.configs`) keep exactly their current local-over-base semantics. No stdout/eval-contract, CLI flag or completion-spec change is involved.

### Trust guard: no silent activation on upgrade

Turning parsed-but-ignored fields into executable configuration is a trust problem: the SHA-256 hash covers a config file's **bytes**, not how vibe interprets them, so a `.vibe.toml` trusted under the old rules could carry a dormant `hooks.post_start_append = ["curl … | sh"]` that starts running on the next `vibe start` purely because the binary was upgraded — with no re-trust prompt anywhere.

This PR versions the interpretation itself and fails closed on the drift:

- `config.rs::CONFIG_SEMANTICS_REV: u32 = 1` names the current interpretation of `.vibe.toml`.
- `AllowEntry` gains an optional `configSemanticsRev` (`settings.rs`), stamped by `add_trusted_path` on both new and re-trusted entries. An absent field means revision 0 — trust granted before this change. `verify_trust_and_read` now returns a `TrustVerdict { trusted, content, semantics_rev }` so the loader gets the revision without re-reading settings (the single-read TOCTOU property is unchanged).
- `config_loader.rs::check_semantics_rev` runs **before** normalization (afterwards the newly-effective positions are indistinguishable from always-effective ones) and rejects a file whose entry is below the current revision *and* which actually uses one of the newly-effective positions: (a) any `*_prepend` / `*_append` in a single-file load, (b) any `*_prepend` / `*_append` in the **base** file of a two-file load, (c) a `*_prepend` / `*_append` set beside its own base field in either file. The error names the file and the exact fields and asks the user to review and re-run `vibe trust`.
- Configs that do **not** use those positions load normally regardless of their entry's revision, and a local-only `_append` wrapping the base array (the one position that always worked) is explicitly not affected — nobody is forced to re-trust for a change that cannot alter their config's meaning.
- The untrusted / hash-mismatch path is unchanged and still wins over the semantics guard.

`configSemanticsRev` is an additive optional field rather than a settings schema bump: neither `AllowEntry` nor `VibeSettings` deserializes with `deny_unknown_fields`, so a v3 reader tolerates it, whereas bumping `version` would push every existing v3 file through the lossy migration ladder to add information those files do not have. `schemas/settings.schema.json` (an editor hint, not runtime validation) is updated to match.

Release notes should still mention the interpretation change and the one-time `vibe trust` some repositories will need.

### Follow-up hardening (review round)

- `vibe verify` now surfaces the guard state: entries below the current revision get a `Config Semantics: trusted at revision N (current: M)` line, and when the file actually uses a newly-effective position, a `⚠️ STALE CONFIG SEMANTICS` verdict naming the exact fields and the `vibe trust` action. The diagnostic and the loader share one predicate (`config_loader::newly_effective_fields` with a `ConfigRole` split), so the advisory output cannot drift from the enforcement. Parse failures are non-fatal for verify: the revision line still prints, the verdict is skipped.
- The "detectors must see the raw, un-normalized config" precondition is now encoded in the type system: `parse_vibe_config` returns a `RawConfig` newtype (its only constructor), the detectors take `&RawConfig`, and `RawConfig::normalize()` is the only way out — a normalized config is not expressible as detector input. (`parse_vibe_config`'s return type changed; no consumer exists outside `vibe-core`.)

## Test Plan

- `rust/crates/vibe-core/src/config.rs` — updated `override_takes_precedence` → `override_replaces_base_then_own_prepend_append_wrap` (deliberate expectation change), plus new `override_with_append_combines`, `empty_override_with_append_yields_append`, `normalize_resolves_same_file_prepend_and_append`, `normalize_append_only_becomes_the_field`, `normalize_resolves_hooks_triplets`, `normalize_is_identity_for_plain_config`, `base_own_append_survives_two_file_merge`, `local_override_replaces_base_effective_array`, and for the guard's input: `reports_no_extension_fields_for_plain_config`, `reports_every_extension_field_in_use_by_dotted_name`, `reports_only_extensions_sharing_their_own_base_field`. Backward-compat pins kept unchanged: `empty_override_array_is_kept`, `copy_files_merging_with_append`, `copy_files_override`, `copy_symlink_merging_with_append`, all scalar-merge tests.
- `rust/crates/vibe-core/src/config_loader.rs` — new `single_base_file_append_is_applied` (the #599 regression test, through the real trust + parse path), `single_base_file_hook_append_without_base_is_applied`, `local_only_append_is_applied`, `base_append_survives_when_local_exists`. Guard tests: `old_rev_entry_without_extension_fields_loads_normally`, `old_rev_single_base_file_with_append_is_rejected`, `old_rev_single_local_file_with_append_is_rejected`, `old_rev_base_extension_with_local_present_is_rejected`, `old_rev_local_field_beside_own_extension_is_rejected`, `old_rev_local_extension_only_still_loads_with_base_present`, `current_rev_entry_makes_extension_fields_effective`, `untrusted_file_with_extension_fields_still_reports_untrusted`. `trusted_base_and_local_are_merged` keeps its assertions.
- `rust/crates/vibe-core/src/settings_io.rs` — `verify_reports_entry_semantics_rev`, `add_trusted_path_upgrades_semantics_rev_of_existing_entry`, plus the golden on-disk entry assertion extended with `configSemanticsRev`; the legacy-entry cases pin that a missing field reads as revision 0.
- Docs updated in all three locales per `.claude/rules/docs-i18n.md`: `configuration/local-config.mdx` (new "Within a Single File" section with the `files = []` + `files_append` caution, and a new "Upgrading: one-time re-trust" section), `configuration/vibe-toml.mdx` (new "Prepend and Append" section plus an upgrade callout), each with `ja/` and `zh/` mirrors; the README triplet carries the same short upgrade note.
- Locally run and passing: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (753 vibe-core tests + integration suites), `pnpm run check:i18n`, `pnpm run fmt:check`, `pnpm run check:docs`.
- Full CI on this PR validates the remaining `check:all` stages (npm shim tests, e2e, licenses).

## Checklist

- [x] Tests added/updated
- [x] `pnpm run check:all` passes (targeted subset run locally; full suite validated by CI on this PR)


kexi:
## Codex review loop — converged (3 rounds)

Reviewer: **OpenAI Codex CLI** (`gpt-5.6-sol`, reasoning effort high, read-only sandbox). Fixes by **Claude**. Round 3 returned zero findings.

### Fixed

- **Trust bypass via the hash history** (round 1, security). `configSemanticsRev` was stamped on the whole allow entry, but `hashes` is a *history* and a match against any past hash grants trust. So re-trusting a benign edit retroactively certified every older hash: a `.vibe.toml` trusted before #599 carrying a dormant `hooks.post_start_append`, later edited and re-trusted, passed the guard once the old revision was checked back out — executing an append never reviewed under the current rules, which is exactly the silent activation the guard exists to prevent. Consent is now recorded per content hash (`configSemanticsRevs`), and `verify_trust_and_read` returns the revision of the hash that actually matched. The `skipHashCheck` path keeps the entry-wide revision (no matched hash exists to attribute consent to), and `semantics_rev_for` falls back to the entry-wide value only for single-hash entries so pre-existing grants are not all downgraded to 0. FIFO-evicted hashes are pruned from the map. Regression test: `retrusting_a_newer_revision_does_not_certify_an_older_hash` (verified to fail without the fix).
- **`vibe verify` drifting from the loader** (round 2, introduced by the fix above). Verify used the per-hash revision unconditionally, so with `skipHashCheck` it could report `STALE CONFIG SEMANTICS` for a file the loader accepts and executes, and on a hash mismatch it printed "trusted at revision N" for bytes carrying no trust at all. The revision is now resolved at the call site mirroring the loader's own branches, and the diagnostic is skipped entirely on a mismatch. Regression tests: `skip_hash_check_reports_the_entry_revision_not_the_on_disk_hash`, `hash_mismatch_reports_no_semantics_revision` (both verified to fail without the fix).

### Disputed

- **`BaseOfPair` over-rejects base-side extension fields** (round 1, minor). Codex argued a base `post_start_append` fully shadowed by a local `post_start` override yields the same result under old and new semantics, so requiring a re-trust breaks the "unchanged meaning keeps working" contract. Rejected: that holds only when the local file *overrides*. When the local *extends* instead (sets `copy.files_append`, not `copy.files`), the base's own append does reach the merged array under the new semantics and did not under the old — see `config::tests::base_own_append_survives_two_file_merge`. `check_semantics_rev` inspects one file at a time and cannot know which case applies, so a fail-closed guard must treat base-side extensions as newly effective; over-rejecting costs one `vibe trust` naming the exact fields, under-rejecting runs an unreviewed hook. Codex accepted this in round 2 and dropped the finding.

### Checks

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (762 vibe-core tests, up from 753), `pnpm run check:i18n` — all passing. `schemas/settings.schema.json` updated for the new key (the entry object is `additionalProperties: false`).

kexi:
@coderabbitai review

PR #607, opened 2026-08-15
fix: emit cd outcome when post_start hooks fail non-fatally

### LINKED ISSUE #601: fix: cd output is lost when a post_start hook fails (exit 0 but no outcome)
## Summary

`HookExecution` errors are designed as "warn and continue" (severity `Warning`, exit code 0), but when `dispatch` returns `Err`, the `Outcome` (the `cd` line) is never written. The user's shell wrapper sees exit 0 with empty stdout, so no `cd` happens even though the worktree was created.

**Location**: `rust/crates/vibe/src/main.rs:72-78` (combined with `?` propagation in `rust/crates/vibe-core/src/commands/start.rs`)

```rust
Err(error) => {
    let exit_code = report_error(&io, &error, quiet);
    if exit_code != 0 { std::process::exit(exit_code); }
}   // exit_code == 0 still ends here → write_outcome is never called
```

## Failure scenario (reproduced via pty)

```toml
# .vibe.toml
[hooks]
post_start = ["exit 3"]
```

`vibe start feat10` → exit code 0, stderr shows `Warning: Hook "exit 3" failed: exit code 3`, stdout is **empty**. The worktree exists (`git worktree list` confirms), but the user stays in the original directory and may believe nothing happened. With `post_start = ["true"]`, the same run prints `cd '/private/tmp/...-feat10'`.

## Background

Not a port regression — the TS normal-start path had the same flaw. However, the TS Claude-Code hook mode explicitly wrapped hooks in try/catch ("post_start hook failures are non-fatal for worktree creation") and continued emitting the path, which matches the declared intent. The current `HookExecution → exit 0` contract contradicts the actual behavior.

## Suggested fix

When the failure is a non-fatal `HookExecution` error after the worktree is created, still emit the `cd` outcome (warn on stderr, cd on stdout).

kexi:
## Summary

`VibeError::HookExecution` is declared warning severity with exit code 0, but returning it from a command still ends the run in `main.rs`'s `Err` arm, which never calls `write_outcome`. A `post_start` hook that exited non-zero therefore produced exit 0 with an **empty stdout**: the worktree existed, the shell wrapper had nothing to eval, and the user stayed in the origin directory believing nothing had happened.

The fix handles the downgrade where the `Outcome` still exists — inside the command layer. A new `hooks::warn_on_hook_failure(io, result) -> Result<bool>` prints the same `Warning: Hook "..." failed: ...` line the binary used to print and swallows the error, returning `false` so an aborting caller can branch.

### `pre_*` gates, `post_*` warn

Which of those two a call site is depends on the hook's **position in the lifecycle**, not on the command. The rule is the same on both sides:

| Hook         | On failure                                                                                     |
| ------------ | ---------------------------------------------------------------------------------------------- |
| `pre_start`  | **Gates**: the copy and `post_start` are already skipped, so no `cd` is emitted either           |
| `post_start` | Warns and still `cd`s — the worktree is fully provisioned. **This is the actual #601 fix**      |
| `pre_clean`  | **Gates**: nothing has been destroyed yet, so the removal is aborted and no `cd` is emitted      |
| `post_clean` | Warns and still `cd`s to main and deletes the branch — the worktree is already gone             |

A failing `pre_*` hook is a failing **safety/precondition** hook, and it must not be followed by the operation it guards. For `clean` that means not destroying the worktree; for `start` it means not walking the user into a worktree the config never finished setting up. Treating `pre_start` like `post_start` (as the first cut of this PR did) would silently break `pre_start` as a precondition check — "is the secrets vault reachable", "is the licence valid" — because the copy and `post_start` are skipped on failure anyway, so continuing hands back an unprovisioned worktree.

Exit status stays 0 in all four rows per the `HookExecution` warning-severity contract. A gated `vibe start` still **creates** the worktree directory (unchanged from before this PR); only entering it is withheld, so the user can fix the cause and re-run `vibe start`.

The gate is **durable**, not one-shot. Because a gated run leaves the worktree on disk, the retry necessarily arrives through a path that hands back an *existing* worktree, so every one of those re-runs the sequence first: creation, same-branch re-entry, `--reuse`, and the "branch is already used in worktree X, navigate?" path. That last one used to return `Outcome::cd(existing)` directly and never call `run_config_and_hooks`, which made the precondition enforceable exactly once and bypassed on every subsequent invocation (and made the "fix the cause and re-run" recovery re-enter without re-provisioning). It now returns an `ExistingBranchDecision::Navigate` **request**, which `start_command` routes through `navigate_to_existing_branch_worktree` once the config is loaded — so the gate re-evaluates, and a cleared precondition finally performs the copy and `post_start` the gated run skipped.

The summary line honours `--quiet`. `warn_on_hook_failure` takes `OutputOptions` and passes it to `format_error_message`, which is where the gating decision lived when `main.rs::report_error(&io, &error, quiet)` still owned the write; moving the write into `vibe-core` must not promote the line to unsuppressable. Only the *printing* is gated — a quiet run is still gated by a failing `pre_*`, and the hook's own stderr (written by `run_hooks`) is unaffected. The two call sites inside `run_config_body` get it via `ConfigAndHooks`, which has no `OutputOptions` of its own.

Coverage of the callers:

- `start`: create, same-branch re-entry, `--reuse` and the interactive Reuse choice all follow the table. `scratch` and `jump`'s create path inherit this through `start_command`.
- A listed submodule's own `pre_start` gates the whole run: its copy and `post_start` were skipped, so the parent worktree is no more usable than if the parent's own gate had failed.
- `start --claude-code-worktree-hook` is untouched: it deliberately ignores the verdict and always emits the path, because Claude Code has no shell to leave in the wrong place.

The match is deliberately narrow — the same call chain also carries `FileSystem` copy failures and `Configuration`/`GitOperation` submodule failures, and widening it would emit a `cd` into a half-built worktree. Only `HookExecution` is downgraded, and tests pin the fatal cases.

Two smaller companion changes:

- Since the warning is now printed from `vibe-core`, the hook command — which comes verbatim from a `.vibe.toml` that is trusted by content **hash**, not by judgement — goes through `output::sanitize_for_display`, so a hostile command string cannot push ESC/bidi sequences onto the terminal (SECURITY_CHECKLIST #6).
- The downgrade happens at the two hook call sites inside `run_config_body`, so `HookExecution` can no longer reach `start::run_config_and_hooks`; its tracker-finish condition collapses to "the run ended non-fatally", which covers the gated path too and keeps a live progress display from being left in front of the prompt. A **fatal** error still leaves the tracker unfinished (issue #600 owns the abort glyph).
- Finishing on the gated path would otherwise have conflated "close the display" with "mark pending tasks succeeded": `IndicatifTracker::finish` closes every still-open bar with the success glyph, so the hooks *after* the failing one — which `run_hooks` never executes — rendered identically to the ones that ran. `ProgressTracker` gains `skip_task`, drawn as `☐ <label> (skipped)`, and `run_hooks` closes the commands it skipped before returning. The same fix covers the pre-existing `post_start` warn-and-cd path.

The eval contract is unchanged: `eval_output.rs::write_outcome` remains the only stdout write point, `main.rs` is not modified, and the paths newly reaching stdout are byte-identical to those the success branches already emit.

Fixes #601

## Test Plan

- Unit (`rust/crates/vibe-core/src/hooks.rs`): the helper downgrades `HookExecution` to the exact warning line and returns `false`; passes `Ok` through; passes **other** errors through unchanged and unprinted; sanitizes ESC/bidi in the hook command; under `--quiet` prints nothing but still reports the failure so the gate still gates; the commands after a failing one are closed as `Skip`, never `Complete`.
- Unit (`rust/crates/vibe-core/src/progress.rs`): `skip_task` is recorded as its own event, distinct from `Complete`.
- Unit (`rust/crates/vibe-core/src/commands/start_tests.rs`): failing `post_start` warns and still cds; failing `pre_start` emits **no** cd, runs no copy and no `post_start`, but still creates the worktree and warns; the **second** run into the worktree the gate left behind re-runs the gate and is gated again, both interactively and under `--force`; once the cause is fixed the re-run runs both hooks, performs the copy and emits the cd; `--reuse` with a failing `pre_start` is gated too; a submodule config's own failing `pre_start` gates the parent run; interactive Reuse and `--reuse` with a failing `post_start` both keep the cd; a submodule **config** error stays fatal at exit 1 with no cd (over-catching guard); tracker `Started`/`Finished` stay balanced on both the warn-and-cd and gated paths, and stay unbalanced on the fatal path.
- Unit (`rust/crates/vibe-core/src/commands/clean_tests.rs`): failing `pre_clean` aborts with no removal and no cd; failing `post_clean` still cds to main and still deletes a configured branch; both hook-mode variants; an untrusted config stays fatal at exit 1.
- Integration (`rust/crates/vibe/tests/eval_contract.rs`, real binary, separate stdout/stderr pipes): `start` with `post_start = ["exit 3"]` exits 0 with stdout **exactly** `cd '<wt>'\n` and the warning on stderr — the direct reproduction of #601; `start` with `pre_start = ["exit 3"]` exits 0 with **empty** stdout while the worktree directory exists, and a second `start --force` for the same branch is gated again rather than cd-ing in; `start --quiet` with a failing `post_start` writes the cd but **no** warning line; `clean` with a failing `post_clean` still writes exactly the cd-to-main line; `clean` with a failing `pre_clean` keeps stdout empty and the worktree on disk.
- Ran locally: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (all green), plus `pnpm run check:i18n` and `pnpm run check:docs`. Full CI validates the rest on this PR.

## Checklist

- [x] Tests added/updated
- [x] `pnpm run check:all` passes (targeted Rust checks + i18n + docs run locally; full CI validates on this PR)
- [x] Docs updated — `packages/docs/src/content/docs/configuration/hooks.mdx` (+ `ja/`, `zh/`) gains a "Hook Failure Behavior" section documenting the `pre_*` gate / `post_*` warn split


kexi:
## Codex review loop — 3 rounds

Reviewer: **Codex CLI (gpt-5.6-sol, reasoning effort high)**, read-only, reviewing `git diff origin/develop...HEAD`. Fixes applied by **Claude**. Each round was a fresh `codex exec` carrying its own context.

### Fixed (5)

- **`start`: `vibe start <branch>` from inside that branch's own worktree re-provisioned itself.** `git rev-parse --show-toplevel` reports the linked worktree, so routing the "branch already used in worktree X" path through `run_config_and_hooks` made `repo_root == existing` — the copy would read and write the same file, and the hooks would guard an entry the user had already made. `navigate_to_existing_branch_worktree` now early-returns the `cd` when `same_worktree(repo_root, existing)`. (`31e22d6`)
- **Hook strings reached the terminal unsanitized on two display paths.** The failure summary was sanitized, but `run_lifecycle_hooks` still echoed the raw command in the dry-run listing and in the progress task labels, in both `start` and `clean` — leaving the SECURITY_CHECKLIST #6 hardening only partly delivered. Only the echo is neutralized; the command still executes byte-for-byte. (`31e22d6`)
- **A hook that could not be *spawned* left its progress bars open.** `runner.run_hook(...)?` skipped the fail/skip bookkeeping; since that error is itself a `HookExecution`, the caller downgrades it and calls `tracker.finish()`, which stamps every open bar with the success glyph — rendering hooks that never ran as completed, exactly what the new `skip_task` exists to prevent. (`31e22d6`)
- **That spawn-failure text was itself unsanitized.** `HookExecution` renders as `Hook "{hook_command}" failed: …`, re-embedding the config-supplied command the task label had just been sanitized for. Now passed through `sanitize_for_display`. (`f3f705d`)
- **`scratch` printed a misleading promote hint after a gated run.** `vibe rename` acts on the *current* worktree, so `Promote with: vibe rename <new-name>` is only correct once the shell is actually moving into the scratch worktree. A gated `pre_start` used to propagate its error and skip the line; downgrading it to `Ok(Outcome::none())` made the hint print while the user was still standing in another branch. Dry-run keeps the hint (no cd by design; the line is part of the preview). (`99eb8f8`)

Each fix carries a regression test, and each was verified to fail against the unfixed code.

### Disputed (2)

- **`start --claude-code-worktree-hook` ignores the `Gated` verdict and emits the path anyway.** `start_worktree_hook_mode` and every `stdout_path` call site are untouched by this PR — this is pre-existing behavior, not a regression. It is also deliberate: the WorktreeCreate protocol hands Claude Code a workspace *path*, and there is no interactive shell to leave in the wrong directory, which is the only harm the gate exists to prevent. The protocol defines no "gated" reply, so the suggested fix would invent an affordance and break the contract.
- **A submodule's copy and `post_start` run before the parent's `pre_start`.** `run_submodule_configs` already ran to completion before the parent's `run_config_body` on `develop`; this PR only adds `Gated` propagation and changes no ordering, so every side effect described occurred identically before it. The contract is per-config — within any single `.vibe.toml`, `pre_start` → copy → `post_start` holds exactly, which is what the docs table claims. The suggested two-phase preflight is a redesign of submodule provisioning, out of scope here.

Codex was given both rejections with this reasoning in round 3 and did not re-raise either.

### Convergence

Not formally converged: the round-3 finding was fixed but the 3-round budget was exhausted before a clean confirming round. The two disputed findings were resolved. Checks green on the final commit: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (756 core tests), `pnpm run check:i18n`.

kexi:
## Review convergence loop — rounds 4-6

Reviewer: **Codex (gpt-5.6-sol, `model_reasoning_effort=high`, read-only sandbox)**. Fixes and triage by **Claude**. Base `origin/develop`, diff `git diff origin/develop...HEAD`.

**Converged**: round 6 returned no findings.

### Fixed (1)

**Round 5 — self-navigation was resolved after `load_vibe_config`** (`rust/crates/vibe-core/src/commands/start.rs`), fixed in 9e074f5.

Making the `pre_start` gate durable moved the "branch is already used in worktree X" answer past the config load, because the `Navigate` request now has to reach `run_config_and_hooks`. The `same_worktree` short-circuit added in round 1 sat *inside* `navigate_to_existing_branch_worktree`, i.e. after that load. So `vibe start feat` run from **inside** the `feat` worktree — which provisions nothing and has no entry to gate — started failing at exit 1 whenever `.vibe.toml` was untrusted or modified. On `develop` that path returned the `cd` before the config was ever read.

The check moved up to the decision site in `start_command`, ahead of the load; the now-unreachable duplicate inside `navigate_to_existing_branch_worktree` was removed rather than left as a second statement of the same rule. Two tests pin the boundary: self-navigation survives an untrusted config, and navigating to a **different** worktree still fails fatally on one — so the short-circuit cannot widen into a way to run an untrusted config's hooks. Both fail with the fix disabled.

### Disputed (1 new, plus 2 carried from rounds 1-3)

**Round 4 — "the parent phase bar (`Pre-start hooks`) is success-stamped by `finish()` even though a child task failed."** Rejected as pre-existing. Phase bars are never explicitly completed *or* failed anywhere in this codebase; `IndicatifTracker::finish` (unchanged in this diff) closes every still-open bar with the success glyph. On `origin/develop` today, `copy_files` warns per-file, calls `fail_task`, returns `()`, and the run reaches `tracker.finish()` — which success-stamps the parent `Copying files` phase over a failed child. A phase-level failure state is a feature request against the tracker, not a defect in a changed line. Re-stated to Codex in rounds 5 and 6 and not re-raised.

**D1 — "`--claude-code-worktree-hook` ignores the Gated verdict."** Rejected as pre-existing and out of diff: `start_worktree_hook_mode` and every `stdout_path` call site are untouched by this PR. Codex accepted the rejection in rounds 4-6.

**D2 — "submodule copy/`post_start` run before the parent `pre_start`, so the gate cannot undo them."** Rejected: `git show origin/develop:.../start.rs` confirms `run_submodule_configs` already ran to completion before the parent hooks on `develop`; this PR did not change that ordering. Accepted in rounds 4-6.

### Follow-up items (separate issues, not this PR)

1. **Phase-level failure rendering.** `ProgressTracker` has no `fail_phase`/`complete_phase`, so a phase whose child task failed still closes with the success glyph. Affects the pre-existing `Copying files` path as much as the hook phases. Natural companion to #600, which already owns the abort glyph for the fatal path.
2. **`--claude-code-worktree-hook` and the gate.** The Claude Code hook mode deliberately always emits the path so the agent is never left in the wrong place. Now that `pre_start` is a real precondition, whether that mode should surface a gated verdict at all is worth deciding on its own merits.

### Checks

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (758 core + 32 + 14 + 13 + 6 + doc tests, all green) and `pnpm run check:i18n` pass locally on 9e074f5. Full CI runs on the PR.

PR #608, opened 2026-08-15
fix: exact-name release asset gate and commit-bound resumable re-runs

### LINKED ISSUE #597: fix: release asset-count gate passes even when a platform binary is missing
## Summary

The release asset gate only checks the **total count** of assets, so a missing platform binary can be masked by an unrelated extra file.

**Locations**: `.github/workflows/release.yml:387`, `.github/workflows/beta-release.yml:215`

## Failure scenario (reproduced in experiment)

1. The `win32-x64` binary upload fails, leaving 4 binaries instead of 5.
2. An extra file appears elsewhere (e.g. 3 `.deb` files are generated, or a stray file lands in `artifacts/` on the beta path).
3. The counts still add up (4+3+2=9 / 4+1+2=7), the gate passes, and a release missing the win32 binary is published.
4. `publish-npm` for `@kexi/vibe-win32-x64` then fails at `gh release download`.

Additionally, glob non-matches leave the literal pattern as 1 array element, and both gate steps lack `set -euo pipefail`, so a failure in the preceding `gh api --paginate` loop is swallowed.

## Suggested fix

Check each expected filename individually with `[ -f ]` (exact expected set per platform) instead of counting, and add `set -euo pipefail` to the gate steps.

Severity: high / reproduced.

### LINKED ISSUE #598: fix: release.yml re-run recovery path is unreachable (should_release=false skips everything)
## Summary

Only the `build` job carries `if: needs.prepare.outputs.should_release == 'true'`; `release` and `update-homebrew` depend on it transitively. On a re-run after a partial failure, `prepare` outputs `should_release=false`, which skips `build` and therefore **everything downstream**, while the run still ends green.

**Locations**: `.github/workflows/release.yml:327`, `:469`

## Failure scenario

1. A release run fails only at `update-homebrew` (e.g. expired tap token).
2. The workflow is re-run.
3. `prepare` sees the release already exists → `should_release=false`.
4. `build` is skipped → `release` and `update-homebrew` are transitively skipped.
5. The Homebrew tap is never updated, and the run is green.

The idempotent "Re-check on re-runs" step inside the `release` job (`release.yml:364`) is only reachable when `should_release=true`, making it dead code (confirmed).

## Suggested fix

Give `release` / `update-homebrew` their own conditions (or make them tolerate skipped `build` via `if: always() && ...` with explicit checks), so a re-run can resume from the failed step idempotently.

Severity: high / confirmed.

kexi:
## Summary

Fixes #597
Fixes #598

Two defects in the release pipeline: one could publish a release missing a platform binary, the other could report a partially-failed release run as green with the failed step never retried.

### #597 — the asset gate counted instead of naming

`release.yml` and `beta-release.yml` built the upload list from a glob and asserted only its **length**. 4 binaries + 3 `.deb` files + 2 license documents also totals 9, so a failed `vibe-win32-x64` upload masked by an unrelated extra file published a release without that binary (`publish-npm` then failed downstream at `gh release download`).

- New `scripts/release-asset-manifest.ts` — the single source of truth for the expected asset set per channel (stable: 5 binaries + 2 version-stamped `.deb`s + 2 license documents; beta: the same minus the `.deb`s), with version validation. `$` matches before a trailing newline in JavaScript, so the anchored regex is paired with an explicit whitespace rejection: a newline in a `.deb` name would split the gate's newline-delimited output into two upload arguments.
- New `scripts/verify-release-files.ts` — pre-upload gate. Verifies every expected file exists as a non-empty **regular** file (`lstat`, so a symlink at an expected path is a problem rather than something to follow — CWE-59) and prints the verified paths, which the workflow feeds straight to `gh release create`. The checked set and the uploaded set are now identical by construction, so the divergence class is removed rather than re-checked. Extra files in `artifacts/` never compensate for a missing one.
- `scripts/verify-release-assets.ts` gains a `--channel/--version` full-set mode requiring exact **set equality** (every expected name present and usable, nothing unexpected attached), which subsumes the old `COUNT` checks. Without the flags its behaviour is unchanged, locked by the pre-existing tests.
- Every gate step gains `set -euo pipefail`, with assignment-capture forms (`DRAFT_IDS=$(...)`, `ASSET_LIST=$(...)`, the SHA256 vars) so a failing `gh api --paginate` / `git ls-remote` / `sha256sum` fails the step instead of reading as an empty result. `for id in $(cmd)` and `echo "$(cmd)"` swallow the exit status even under `set -e`; those shapes are load-bearing, not stylistic.

### #598 — the re-run recovery path was unreachable

Only `build` carried `if: needs.prepare.outputs.should_release == 'true'`. On a re-run after a partial failure (e.g. an expired tap token), `prepare` reported `should_release=false`, `build` was skipped, and `release` / `update-homebrew` were skipped transitively — the run ended green with the Homebrew tap never updated. The idempotent "Re-check on re-runs" step inside `release` was dead code.

- `release` and `update-homebrew` now carry explicit `if:` conditions using `!cancelled()` plus per-need result checks, so a resume runs them: the re-check no-ops before any create/delete/upload, the post-publish verify re-asserts the full published asset set, and the tap converges via the existing `git diff --cached --quiet` short-circuit. `!cancelled()` overrides the implicit needs gate, so every transitive dependency's result is checked explicitly — a cancelled run or a failed `prepare` still skips everything.
- **Commit binding (security review, REQUIRED)**: those jobs `actions/checkout` the default ref, which on a *fresh* dispatch is main's current HEAD, not the released commit. That would hash the already-released binaries into a formula built from a newer tree (new `Formula/*.rb`, new `assert-formula-license.sh`) and push it to the tap. `prepare` therefore compares the published release's `targetCommitish` to `GITHUB_SHA` and fails closed with an `::error` telling the operator to use "Re-run failed jobs". Mirrored in `beta-release.yml`.
- `update-homebrew` gates on a new `verified` job output set by the post-publish asset check, so that check is load-bearing for the tap rather than incidentally reachable (on a resume, `release` succeeds via its early `exit 0`).
- `update-homebrew` sources its binaries from the published release rather than build artifacts — the bytes users actually install, and the only source available on a resume — by **exact** `--pattern` names rather than `vibe-darwin-*` globs, matching the #597 discipline one layer down.
- `beta-release.yml` gets the same published re-check guard in its release job, making "Re-run all jobs" (same `run_number`, hence same tag) resumable.

No third-party action `uses:` lines were added or changed (`pinact run --check` passes), no new secrets: `gh release download` uses the existing `github.token`, and `HOMEBREW_TAP_TOKEN`'s scope is untouched.

## Test Plan

- New `packages/npm/test/release-asset-manifest.test.ts` (14 tests): the exact stable/beta sets and their sources, version rejection for prerelease-on-stable, malformed versions, shell metacharacters, the trailing-newline case the anchored regex alone accepts, an unknown channel, and the no-whitespace-in-names invariant the `mapfile` protocol depends on.
- New `packages/npm/test/verify-release-files.test.ts` (15 tests): the #597 regression (missing `vibe-win32-x64` plus a stray extra reports exactly one problem naming the binary), zero-byte file, directory at an expected path, symlink rejection, `LICENSE` looked for in the checkout rather than `artifacts/`, path resolution order, and CLI arg parsing.
- Extended `packages/npm/test/verify-release-assets.test.ts`: full-set mode where a stray replaces the win32 binary at an unchanged count of 9 reports both the missing and the unexpected asset; size/state now guard binaries and `.deb`s, not just the license documents; `--channel`/`--version` must be given together; legacy no-flag behaviour still locked by the original tests.
- `pnpm run test:npm` — 287 tests pass across 13 files.
- `actionlint` on both workflows: no expression or syntax findings; the shellcheck info count dropped from 14 to 3, and the remaining 3 are pre-existing lines this PR does not touch.
- `pinact run --check`, `pnpm run fmt:check`, `pnpm run lint`, `pnpm run check:i18n` pass.
- Both gate CLIs smoke-tested end to end against a staged fixture tree, including the `ASSET_LIST=$(...)` / `mapfile` round-trip under `set -euo pipefail`.

Full CI validates this PR. There is no runnable harness for job-level `if:` graphs, so the enforcement is the extracted, tested gate scripts; a post-merge `gh workflow run release.yml -f dry_run=true` on main would additionally rehearse the file gate, full-mode asset verification and draft cleanup end to end.

## Checklist

- [x] Tests added/updated
- [x] `pnpm run check:all` passes (targeted checks run locally; full CI validates on this PR)
- [ ] Docs updated (if needed) — no user-facing behaviour change; rationale lives in the workflow comments


kexi:
## Codex review loop — 3 rounds

External reviewer: **Codex (gpt-5.6-sol, reasoning effort high)**, read-only, reviewing `git diff origin/develop...HEAD`. Fixes applied by **Claude**.

### Fixed (6)

- **Resumable `release` let a re-run of an OLD run roll the Homebrew tap back** (R1, both workflows). `prepare`'s commit binding compares the release to the re-run's *own* commit, which still matches, so re-running the 3.1.0 run after 3.2.0 shipped would have rewritten the unversioned `Formula/vibe.rb` back to 3.1.0 and pushed it — the `git diff --cached --quiet` short-circuit does not help, since the regenerated old formula genuinely differs. `update-homebrew` now gates every tap step on the release still being current, skipping with a `::warning::` rather than failing so an unrelated re-run is not blocked. — `2981d96`
- **`isLatest` is not a `gh release view` field** (R2). It belongs to `gh release list`; the guard would have failed on *every* stable release, taking `update-homebrew` down with it. Now queried via `gh release list`. — `d0570d2`
- **Both guards listed releases without `--exclude-drafts`** (R2). An authenticated listing returns this repo's drafts, and both workflows create their release as a draft first, so a newer run that died before undrafting would have read as "something newer shipped" and skipped the repair the re-run exists to perform. — `d0570d2`
- **`--limit 1` applied before the jq filter** (R2, found while fixing the above). One non-matching entry at the head would yield an empty result and skip the tap on a genuinely current release. Both guards now fetch a window and filter in jq. — `d0570d2`
- **`parseCliOptions` accepted a valueless `--version`** (R2, `scripts/verify-release-assets.ts`). It stored `undefined`, which the "given together" check then read as *neither* flag present — silently serving the weak license-only mode to a caller asking for full-set verification. Now throws, with a test. — `d0570d2`
- **Beta guard ordered by release creation time** (R3). `gh release list` sorts by creation time and a re-run recreates its release, so a recreated old beta would carry a fresh timestamp and sort first, walking through the guard. Now compares the `-beta.N` ordinal (`github.run_number`, monotonic) numerically. Also corrected `CONTRIBUTING.md`, which still said "Re-run all jobs" skips the pipeline once published — now the opposite of what the workflow does. — `5190879`

### Disputed (2)

- **"Re-running an old *unpublished* run republishes it as `--latest` and rolls back the tap"** (R3, `release.yml`) — real, but **pre-existing and not introduced here**. That path goes through `should_release=true`, which is unchanged by this PR: on `origin/develop` the same re-run already builds, publishes with `gh release edit --draft=false --latest`, and reaches `update-homebrew` (gated only on `inputs.dry_run != true`, with `release` succeeded). Identical outcome before and after. The regression this PR *did* introduce — the already-published resume path — is fixed above. The same reasoning covers the beta variant.
- **"npm `latest` dist-tag could roll back via `publish-npm.yml`"** (R3) — out of scope. `publish-npm.yml` is not touched by this branch (`git diff origin/develop...HEAD -- .github/workflows/publish-npm.yml` is empty), and the `workflow_run` re-fire path it depends on predates this PR for the same reason as above.

### Convergence

Not fully converged: round 3 returned two findings, both disputed above as pre-existing/out-of-scope rather than defects in this diff. Every finding scoped to code this PR changes was accepted and fixed. Reached the 3-round limit.

### Checks

`pnpm run test:npm` 288 passed (13 files) · `actionlint` no expression or syntax findings, only the 3 pre-existing shellcheck infos this PR does not touch · `pnpm run fmt:check`, `pnpm run lint`, `pnpm run check:i18n` pass.

kexi:
## Review convergence loop — rounds 4–6

Reviewer: Codex (`gpt-5.6-sol`, reasoning effort `high`, read-only sandbox). Fixes by Claude.
Base for every round: `git diff origin/develop...HEAD`.

**Status: converged.** Rounds 5 and 6 both returned zero findings.

### Findings fixed (1)

| Round | Severity | Finding | Fix |
| --- | --- | --- | --- |
| 4 | minor | `parseCliArgs` in `scripts/verify-release-files.ts` accepted a valueless or empty `--repo-root` and silently fell back to `.` | `ac56930` — throws `--repo-root requires a value`, plus two tests |

Why it mattered: the flag names the checkout the license documents are read from. Left `undefined`, the default read it as "the flag was never given" and resolved `LICENSE` / `THIRD-PARTY-LICENSES.md` against the process CWD instead — so the gate could report OK for files other than the ones about to be uploaded. Same silent-downgrade class already closed for `--version` in `verify-release-assets.ts` in round 3, so closing it here keeps the two gates consistent.

### Disputes re-litigated

Both earlier disputes were put back to Codex in round 4 with the rejection evidence, and it **explicitly accepted both** ("D1・D2の却下はいずれも妥当で、差分による新たな到達可能性や悪化はありません" — both rejections are sound; the diff introduces no new reachability and makes neither worse). Neither was re-raised in rounds 5 or 6.

- **D1** — "re-running an old *unpublished* run republishes with `--latest` and rolls the tap back". Real hazard, but that path runs through `should_release=true`, which this branch does not change; identical on `origin/develop`.
- **D2** — "npm `latest` dist-tag rollback via `publish-npm.yml` `workflow_run` re-fire". `git diff origin/develop...HEAD -- .github/workflows/publish-npm.yml` is empty; the re-fire path predates this PR.

### Follow-up items (out of scope here, worth separate issues)

Both are pre-existing on `develop` and were deliberately not fixed in this PR to keep its scope on #597/#598:

1. A re-run of an old *unpublished* release run can still publish that old version with `--latest` and roll `Formula/vibe.rb` backwards. The tap-rollback guard added in this PR covers the resumed-published path; the `should_release=true` path is untouched.
2. `publish-npm.yml` can re-fire via `workflow_run` and move the npm `latest` dist-tag back to an older version.

### Rounds 4–6 totals

- Rounds run: 3 (round 4: 1 finding; round 5: clean; round 6: clean)
- Fixed: 1 · Disputed: 0 new (2 prior disputes re-confirmed by the reviewer)
- `pnpm run test:npm`: 290 passed / 13 files · `pnpm run fmt:check`: clean

PR #609, opened 2026-08-15
fix: handle EPIPE in verify-deb tar pipeline

### LINKED ISSUE #602: fix: verify-deb.ts crashes with unhandled EPIPE instead of reporting the verification error
## Summary

`dpkg.stdout.pipe(tar.stdin)` has no handler for the case where tar exits first; the resulting `EPIPE` on `tar.stdin` crashes Node with a raw stack trace instead of the intended verification error.

**Location**: `scripts/verify-deb.ts:161`

## Failure scenario (reproduced)

1. The .deb lacks `usr/share/doc/vibe/copyright`.
2. tar exits immediately with code 2, while dpkg-deb keeps streaming ~78MB.
3. The write to `tar.stdin` throws an unhandled `EPIPE`; Node aborts with a stack trace instead of the intended `could not extract ... (tar exit 2)` message, making the release-gate failure reason undiagnosable.

## Also

The comment at `verify-deb.ts:182-183` ("tar exits 0 when asked for a member that does not match anything") is factually wrong — both GNU tar and bsdtar exit 2. The code is safe (`tarCode !== 0`), but the comment states the opposite premise.

## Suggested fix

Handle `error` on `tar.stdin` (ignore `EPIPE`), let the tar exit-code check report the failure, and fix the comment.

kexi:
## Summary

`scripts/verify-deb.ts` wired `dpkg-deb --fsys-tarfile` into `tar -xO` with a bare
`dpkg.stdout.pipe(tar.stdin)`. When tar left before draining its input, the write to its
stdin emitted `EPIPE` with no listener attached and Node aborted the whole verifier with a
raw stack trace — so the release gate failed with an unhandled stream error instead of the
intended `could not extract usr/share/doc/vibe/copyright from the package (tar exit 2)`.

The pipe wiring is now extracted into `pipeIgnoringBrokenPipe(source, sink)` (exported and
unit-tested), with `isBrokenPipeError()` classifying the codes raised when a pipe's consumer
is gone.

Three things had to be handled, not one:

1. **Swallow the broken pipe.** `EPIPE` / `ERR_STREAM_DESTROYED` /
   `ERR_STREAM_WRITE_AFTER_END` on either end are ignored, so the existing exit-code checks
   do the reporting. Any other stream error is left unhandled, so a real I/O fault is not
   silently turned into an empty extraction.
2. **Tear the source down.** `pipe()` unhooks itself as soon as the sink goes away, so
   nothing drains dpkg-deb any more and it blocks forever writing ~78 MB into a full pipe —
   handling only the error would have converted the crash into a hang. Destroying the read
   end makes the producer see the broken pipe and exit, which is what lets both `close`
   events arrive.
3. **Watch `close` as well as `error`.** Whether a departing consumer shows up as an EPIPE
   on the next write or as a silently closed stdin is a race; handling only the error path
   left the hang reachable roughly half the time (reproduced: 1 in 2 runs hung locally).

`tar`'s status is now checked before `dpkg-deb`'s: once tar leaves early, dpkg-deb is killed
by SIGPIPE (exit `null`) purely as a consequence, so reporting the producer would name the
wrong process and hide the real reason. The corrupt-archive guard on `dpkgCode` is unchanged
and still fires.

Also fixes the factually wrong comment at the old `:182-183`: tar does **not** exit 0 for a
member that matches nothing (GNU tar exits 2, bsdtar exits 1 — both verified locally). The
comment is replaced with what the emptiness check actually guards, namely a member that is
present but empty.

## Test Plan

**Unit tests** — 5 new cases in `packages/npm/test/verify-deb.test.ts` (the existing vitest
harness for this script), driving the pipe with stand-in `node -e` processes:

- `isBrokenPipeError` recognises all three broken-pipe codes and rejects unrelated errors
  (`ENOSPC`, a plain `Error`, `undefined`).
- A consumer exiting **2** mid-stream: both exit codes are readable, no crash, no hang.
- A consumer exiting **0** mid-stream (the `tar --fast-read` shape): same teardown applies.
- Happy path: every byte is forwarded.
- An 8 MB payload is forwarded untruncated — guards the teardown from firing on a healthy
  stream, which would cut a multi-megabyte member short.

Verified the crash test genuinely regresses: reverting to the bare `pipe()` makes it fail
with the exact `Error: write EPIPE` / `Unhandled 'error' event` trace from the issue. Ran
`test/verify-deb.test.ts` 6× consecutively to confirm the race is gone (29/29 each run).

**End-to-end against a real archive.** `dpkg-deb` does not exist on macOS, so I built a
fixture and a stub: an 80 MB uncompressed `data.tar` containing `./usr/bin/vibe` (80 MB,
mode 755), a DEP-5 `./usr/share/doc/vibe/copyright` and
`./usr/share/doc/vibe/THIRD-PARTY-LICENSES.md` with the real markers, plus a `dpkg-deb`
shim implementing `--contents` (GNU `tar -tvf`, which produces the same listing layout dpkg
does) and `--fsys-tarfile` (`cat`). The script was then run unmodified under **both** `bun`
(as CI does) and `node --experimental-strip-types`:

| Scenario | Stub | Result |
| --- | --- | --- |
| A. Complete package | real GNU tar | `✓ work.deb: ships usr/bin/vibe, …` exit 0 |
| B. tar exits 2 before draining stdin | `tar` stub: `head -c 65536 >/dev/null; exit 2` | `verify-deb: could not extract usr/share/doc/vibe/copyright from the package (tar exit 2)` exit 1 |
| C. dpkg-deb streams fully then exits 2 | `dpkg-deb` stub | `verify-deb: dpkg-deb could not read the package (exit 2)` exit 1 |
| D. dpkg-deb fails on a corrupt stream | `dpkg-deb` stub exits 2 with no payload | `verify-deb: could not extract … (tar exit 2)` exit 1 |

Scenario B on the pre-fix code under Node reproduces the issue exactly:

```
tar: giving up before draining stdin
node:events:487
      throw er; // Unhandled 'error' event
Error: write EPIPE
    at WriteWrap.onWriteComplete [as oncomplete] (node:internal/stream_base_commons:87:19)
  errno: -32, code: 'EPIPE', syscall: 'write'
```

Worth recording: under **bun** (the runtime CI uses) the pre-fix code happens not to crash —
bun's stream polyfill absorbs the EPIPE. That is an implementation detail rather than a
guarantee, and the vitest suite runs under Node where the crash is real, so the fix makes
the behaviour runtime-independent rather than relying on bun.

**Checks run locally:** `pnpm run lint`, `pnpm run fmt:check`, `pnpm run test:npm`
(249/249 pass). Full `check:all` (Rust, e2e, docs, i18n, licenses) is validated by CI on
this PR.

## Checklist

- [x] Tests added/updated
- [x] `pnpm run lint` / `fmt:check` / `test:npm` pass (full `check:all` runs in CI)
- [ ] Docs updated (not needed — internal release-gate script)

Fixes #602


kexi:
## Review convergence loop: 2 rounds, converged

Reviewer: **Codex** (`gpt-5.6-sol`, reasoning effort high). Fixes by **Claude**.

### Round 1 — 1 finding, accepted and fixed

**`scripts/verify-deb.ts` — a source-side I/O fault hung the verifier instead of reporting itself** (major)

The `source` `error` handler recorded the error but only destroyed the source. `pipe()` forwards `end` to the sink only on a clean end-of-stream, so a destroyed source left `tar.stdin` open and tar waited for an EOF that never arrived. Since `readMember` reads `pipeFailure()` only *after* awaiting both `close` events, the recorded fault was never surfaced at all — the release gate hung rather than printing `verify-deb: failed while streaming the package: ...`.

Reproduced under Node before the fix (consumer never closed, 2 s budget exhausted); passes after. Fixed in 43115ad by destroying the sink from the source's `error` handler. The sink is destroyed rather than `end()`ed so tar cannot parse a truncated member and report a plausible-looking "marker missing" verdict in place of the real I/O error.

Codex also noted the existing test masked the bug: `hands a genuine I/O fault back to the caller` killed both processes before checking the getter. It now awaits the consumer's own `close`, exactly as `readMember` does, so the hang fails the test instead of hiding in it. Verified the test times out without the production fix and passes with it.

### Round 2 — no findings

Codex re-verified the fix and re-reviewed the whole diff, including running the early-exit and fault paths 25 iterations each (no flakes) and confirming the 8 MB healthy-stream case still forwards every byte with both processes exiting 0.

### Disputed

None — the single finding was genuine.

### Checks

`pnpm run lint`, `pnpm run fmt:check`, `pnpm run test:npm` all pass (254 tests).

kexi:
@coderabbitai review

PR #619, opened 2026-08-16
fix: resolve hoisted node-pty before chmodding spawn-helper in e2e postinstall

### LINKED ISSUE #618: fix: e2e postinstall chmod misses hoisted node-pty spawn-helper (silenced by || true)
## Summary

`packages/e2e`'s `postinstall` runs `chmod +x node_modules/node-pty/prebuilds/*/spawn-helper` with a path relative to the package, but pnpm hoists `node-pty` to the **root** `node_modules`, so the glob matches nothing. The trailing `|| true` swallows the failure silently. On an affected checkout the whole e2e suite fails with `posix_spawnp failed` (reproduced 71/71 failures on unmodified develop).

## Suggested fix

Resolve the real node-pty location (e.g. `require.resolve('node-pty/package.json')` in a small script) before chmod, and drop the `|| true` so a miss is loud. Alternatively use pnpm's `dependenciesMeta`/supported lifecycle handling instead of a manual chmod.

Found while validating PR #617 (Dependabot alert fixes); pre-existing on develop, not introduced there.

Found by automated dependency-fix audit (Claude Code).

kexi:
## Summary

`packages/e2e`'s `postinstall` existed to restore the execute bit on node-pty's
prebuilt `spawn-helper` (the tarball ships it `0644`, and the addon
`posix_spawnp`s that path for every PTY it opens). It globbed
`node_modules/node-pty/prebuilds/*/spawn-helper` **relative to `packages/e2e`** —
but `.npmrc` sets `node-linker=hoisted` with `public-hoist-pattern[]=*node-pty*`,
so node-pty actually installs into the **workspace root** `node_modules`. The
glob matched nothing and the trailing `|| true` swallowed the miss.

### Reproduced (on unmodified `develop`)

```
$ rm -rf node_modules packages/*/node_modules && pnpm install --frozen-lockfile
packages/e2e postinstall$ chmod +x node_modules/node-pty/prebuilds/*/spawn-helper 2>/dev/null || true
packages/e2e postinstall: Done          # looks clean

$ ls -d packages/e2e/node_modules/node-pty
ls: cannot access 'packages/e2e/node_modules/node-pty': No such file or directory

$ find node_modules -name spawn-helper -exec ls -la {} +
-rw-r--r--  node_modules/node-pty/prebuilds/darwin-arm64/spawn-helper   # no +x
-rw-r--r--  node_modules/node-pty/prebuilds/darwin-x64/spawn-helper     # no +x

$ pnpm -C packages/e2e test
Error: posix_spawnp failed.
Test Files  11 failed (11)
     Tests  71 failed | 1 skipped (72)
```

### Fix

New `scripts/fix-node-pty-spawn-helper.ts` (dependency-free — `node:fs`,
`node:module`, `node:path` only) replaces the shell one-liner:

- locates node-pty via `require.resolve("node-pty/package.json")` from the e2e
  package context — the same lookup node performs at test time, so it tracks the
  layout rather than guessing at it;
- `chmod`s every `prebuilds/*/spawn-helper` found and **prints what it fixed**;
- **exits non-zero** when node-pty resolves but no `spawn-helper` is found. The
  `|| true` is gone. The loud exit is the point: a silent miss is precisely how
  this survived unnoticed;
- skips cleanly (exit 0, with a message) on `win32`, where node-pty uses
  ConPTY and no `spawn-helper` exists;
- treats an absent node-pty as a no-op, so `--filter`ed / `--ignore-scripts`
  installs stay quiet.

Note this only ever bit local development: CI installs with `--ignore-scripts`
and rebuilds node-pty from source, so the e2e job was never affected.

## Test Plan

- **The real proof** — fresh `pnpm install --frozen-lockfile` (exercising the new
  postinstall), then the actual e2e suite:

  ```
  packages/e2e postinstall$ bun run ../../scripts/fix-node-pty-spawn-helper.ts .
  packages/e2e postinstall: fix-node-pty-spawn-helper: chmod +x .../prebuilds/darwin-arm64/spawn-helper
  packages/e2e postinstall: fix-node-pty-spawn-helper: chmod +x .../prebuilds/darwin-x64/spawn-helper

  $ pnpm run test:e2e
  Test Files  11 passed (11)
       Tests  71 passed | 1 skipped (72)
  ```

  71/71 failing → 71/71 passing.

- Added `packages/npm/test/fix-node-pty-spawn-helper.test.ts` (14 cases) to the
  existing vitest suite — no new test infrastructure. It pins the behaviour that
  regressed: node-pty **hoisted to the workspace root** is found from a nested
  package, a package-local copy still wins, a missing node-pty yields `null`
  instead of throwing, `prebuilds/` enumeration tolerates helper-less
  directories (win32) and an absent `prebuilds/` while reporting the empty list
  the caller fails loudly on, and the execute-bit predicate requires all three
  bits (what makes the chmod idempotent).

- Verified the loud-failure path fires: a node-pty install with no
  `prebuilds/` exits 1 with a diagnostic naming the resolved directory.

- Verified idempotency: a second run reports "already executable" and exits 0.

Checks run and passing:

- `pnpm install --frozen-lockfile` (fresh, postinstall exercised)
- `pnpm run test:npm` — 311 passed (14 files)
- `pnpm run test:e2e` — 71 passed, 1 skipped
- `pnpm run check:rust`, `fmt:check`, `lint`, `check:i18n`, `check:licenses`

## Checklist

- [x] Tests added/updated
- [ ] Docs updated (if needed) — n/a, internal build tooling only

Fixes #618


kexi:
## Review–fix convergence loop: 3 rounds, 5 findings, all fixed

Reviewer: **Codex (gpt-5.6-sol, reasoning effort high)** — fresh `codex exec` each round, no resume.
Fixes and verification: **Claude**.

Every finding was triaged against the actual node-pty sources in `node_modules/` and reproduced before being accepted. **No finding was disputed** — all five were genuine, and three of them would have shipped a broken `pnpm install`.

### Round 1 — 3 findings, all valid

| # | Severity | Finding |
|---|----------|---------|
| 1 | **critical** | The loud `exit(1)` fired on every platform, but node-pty builds `spawn-helper` on **macOS only** |
| 2 | major | Running the script with `bun` made bun a hard requirement of `pnpm install` |
| 3 | minor | `build/Debug` was not probed, though node-pty's loader searches it |

**#1** was the serious one. `binding.gyp` declares the `spawn-helper` target under `['OS=="mac"']`, and `src/unix/pty.cc` reads `helperPath` solely inside `#if defined(__APPLE__)` — every other Unix forks and `execvp`s directly. The tarball also ships no `prebuilds/linux-*`, so Linux **always** compiles from source and legitimately ends up with no helper at all. Reproduced against a simulated Linux from-source tree: `helpers found: []` → `would exit(1): true`. As written, this would have broken `pnpm install` on every Linux dev machine and in Docker, on a perfectly healthy tree. Now gated behind `helperIsRequired(platform)`.

**#2**: the branch had added a CONTRIBUTING.md note telling contributors to install from inside the dev shell "or the install fails with `bun: command not found`" — documenting the regression rather than fixing it. The script is now plain `.mjs` run by `node` (guaranteed by `engines.node`), the note is gone, and the bun-only `import.meta.main` guard became an `argv[1]` comparison. JSDoc types rather than TS syntax, since type stripping is unavailable on the supported node floor (>=18).

### Round 2 — 1 finding, valid

The macOS guard accepted the tree whenever **any** helper was found. But node-pty doesn't search for the helper independently: `loadNativeModule()` picks the first directory holding a loadable `pty.node`, and `unixTerminal.js` derives `helperPath` from *that same directory*. So on an arm64 host missing only `prebuilds/darwin-arm64/spawn-helper`, the leftover `darwin-x64` helper satisfied the count — postinstall reported success while E2E still died with `posix_spawnp failed`, exactly the #618 symptom. Confirmed by deleting that one helper: the old check passed, the new one exits 1 naming the exact path. Verdict now comes from `activeSpawnHelper()`; repair still chmods every helper found, since fixing an inactive one is harmless.

### Round 3 — 1 finding, valid

`activeSpawnHelper()` tested file *existence*, but node-pty `require()`s each candidate and **falls through when the load throws**. A stale `build/Release/pty.node` from a wrong-arch checkout or half-finished compile still exists while failing to dlopen — so node-pty skips it and runs the prebuild's helper, and an existence-only check pinned the verdict to the wrong directory in either direction. The probe now mirrors node-pty by attempting the require. Verified safe for postinstall: dlopen failures surface as ordinary catchable `Error`s and the process survives, checked against both a corrupt file and a real wrong-arch `pty.node`. The loader is injectable so tests stay hermetic, with one test on the default argument proving the real probe rejects placeholders without crashing.

### Verification

Note that **CI green is not evidence here** — the workflows install with `--ignore-scripts` and rebuild node-pty from source, which is precisely why #618 survived. The real proof is a fresh install plus a full e2e run:

- `pnpm install --frozen-lockfile` — postinstall runs under `node`, reports helpers already executable
- helpers reset to `0644` → re-run restores `0755` (the actual #618 repair, end to end)
- only `prebuilds/darwin-arm64/spawn-helper` removed → exits 1; healthy tree → exits 0
- `pnpm run test:npm` — 327 passed (13 → 26 in this suite)
- `pnpm run test:e2e` — 71 passed, 1 skipped
- `pnpm run fmt:check`, `pnpm run lint` — clean

### Convergence

Reached the 3-round cap with each round surfacing one genuine, narrower issue than the last — rounds 2 and 3 both refined the same `activeSpawnHelper` predicate rather than finding new problem areas. Nothing outstanding, no disputed findings, and no follow-ups recorded. Platform caveat worth stating plainly: all runtime verification was on **macOS arm64**, which is also the only platform where this guard is armed; the Linux and Windows paths are covered by unit tests over `helperIsRequired`, not by execution on those hosts.

PR #620, opened 2026-08-16
feat: signal a gated pre_start to Claude Code hook mode on stderr

### LINKED ISSUE #615: design: should --claude-code-worktree-hook surface a gated pre_start verdict?
## Summary

PR #607 makes a failing `pre_start` hook gate the `cd` for interactive `vibe start` (mirroring `pre_clean`). The `--claude-code-worktree-hook` mode deliberately ignores the `Provisioning::Gated` verdict and always emits the worktree path, because Claude Code has no shell to strand.

Now that `pre_start` is a real precondition gate, decide whether hook mode should surface the gated state to the caller (e.g. a marker in output, a non-zero exit, or a structured field) instead of silently handing over an unprovisioned worktree.

Pre-existing behavior, out of #607's scope by design (the mode's call sites are untouched there). Raised by Codex review of PR #607.

Found during the Codex/Claude review convergence loop.

kexi:
## Summary

Since #607 a failing `pre_start` is a real precondition gate for interactive
`vibe start`, but `--claude-code-worktree-hook` mode dropped the
`Provisioning::Gated` verdict and handed back an unprovisioned worktree that
looked identical to a good one. The only trace was the generic
`Warning: Hook "..." failed: ...` line, which embeds the user's hook command and
is therefore not a signal a caller can key on.

Per the decision on #615, this keeps the hook-mode contract intact and adds a
dedicated signal instead:

- **stdout is unchanged**: the worktree path, verbatim, no `cd`.
- **exit status is unchanged**: 0.
- **stderr gains one fixed line** when — and only when — a `pre_start` gate
  fired:

  ```
  vibe: pre_start hook failed; worktree is not provisioned
  ```

A non-zero exit was rejected because it would make the caller read an existing
worktree as a failed creation, whose recovery story is worse than knowingly
entering an unprovisioned one. This mirrors #607's interactive semantics: humans
are not `cd`'d into an unprovisioned worktree, agents get the path plus an
explicit signal.

Implementation notes:

- The text lives in a single constant,
  `vibe_core::commands::start::HOOK_MODE_GATED_SIGNAL`, so the unit tests, the
  real-binary integration test, and the implementation cannot drift.
- It is written straight through the `Io` seam rather than via `warn_log`,
  because `warn_log` wraps its message in ANSI yellow when color is enabled,
  which would defeat a byte-exact match by a consumer.
- Emitted at most once per invocation (only `pre_start` can gate, and the
  config-and-hooks sequence runs once), on both the creation path and the
  same-branch re-entry path.
- A failing `post_start` is unaffected: the worktree is fully provisioned, so it
  still only produces the generic warning.

A structured verdict (a JSON field Claude Code could consume) is deliberately
left out of scope, as noted on the issue.

## Test Plan

New unit tests in `rust/crates/vibe-core/src/commands/start_tests.rs`:

- `worktree_hook_mode_gated_pre_start_emits_the_signal_and_still_outputs_the_path`
  — stdout is the path, no `cd`, the worktree is created, copy and `post_start`
  are skipped, and the signal appears on stderr **exactly once** as a bare line.
- `worktree_hook_mode_successful_pre_start_emits_no_gated_signal` — a fully
  provisioned run stays silent, so the signal is unambiguous.
- `worktree_hook_mode_failing_post_start_emits_no_gated_signal` — a `post_start`
  failure warns but is never reported as a gate.

New real-binary integration test in `rust/crates/vibe/tests/eval_contract.rs`:

- `start_worktree_hook_mode_gated_pre_start_signals_on_stderr` — spawns the
  actual binary with separate stdout/stderr pipes and asserts exit 0, the bare
  path on stdout, and exactly one matching stderr line.

Existing hook-mode tests (`worktree_hook_mode_post_setup_failure_is_non_fatal`,
`worktree_hook_mode_outputs_path_to_stdout`, and the rest) are unchanged and
still pass.

Checks run: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`,
`cargo test --workspace` (1037 + 41 + 13 + ... all green), `pnpm run check:i18n`,
`pnpm run fmt:check`, `pnpm run check:docs`.

## Checklist

- [x] Tests added/updated
- [x] `pnpm run check:all` passes (fmt / clippy / workspace tests / i18n / docs run individually)
- [x] Docs updated (`configuration/hooks.mdx` + `ja` / `zh` mirrors)

Fixes #615


kexi:
## Review convergence loop — converged after 2 rounds

Reviewer: **Codex (gpt-5.6-sol, reasoning effort high)** · Fixes and triage: **Claude**

Each round ran as a fresh `codex exec` over `git diff origin/develop...HEAD` with no carried context; round 2 was given the round-1 disputes plus their rejection evidence for explicit re-evaluation.

### Round 1 — 2 findings

**Fixed — docs contradicted their own setup example** (minor)
The new "Claude Code hook mode" section told readers to key on a fixed stderr line, while the recommended `settings.json` on the same page ends the command with `2>/dev/null` — so anyone following the setup verbatim would never see the signal. All three locales now carry a caveat pointing at the setup example and telling the reader to drop the redirection.

**Fixed in part / disputed in part — "the `SameBranch` signal handling is unreachable"** (major)
The facts checked out: `find_worktree_by_branch` and `get_worktree_by_path` both parse the same `git worktree list --porcelain`, so a worktree already on `branch_name` is always caught by the `!validation.is_valid` early return, and the `ConflictType::SameBranch` arm in hook mode is effectively unreachable. Re-running the hook after a gated run therefore emits no signal.

Disputed as a defect *of this diff*: both the early return and the `SameBranch` arm exist verbatim on `origin/develop`, which likewise runs no hooks on that path — the diff only added a signal inside a pre-existing arm. Routing the early return through `run_config_and_hooks` would be a behavior change beyond this PR's scope (the config is not loaded at that point) and would break the pre-existing `worktree_hook_mode_existing_branch_outputs_existing_path_without_creating`.

What *was* wrong and got fixed: the new code comment claimed that arm was the re-entry path. It is not. The comment now records the real boundary, the limitation is documented as a caveat in all three locales, and `worktree_hook_mode_reentry_on_existing_worktree_emits_no_gated_signal` pins the behavior so the docs and the code cannot drift. The underlying re-entry gap predates this change and is left for separate follow-up.

### Round 2 — 0 findings

Codex re-examined the disputed finding against the evidence above and returned `[]`. Converged.

### Checks

`cargo fmt --all --check` · `cargo clippy --workspace --all-targets -D warnings` · `cargo test --workspace` (1119 passed, 0 failed) · `pnpm run check:i18n` — all green.

Issue #611, opened 2026-08-15
fix: post_start phase spinner is left frozen because finish() is never reached


kexi:
## Summary

Found during review of PR #605: on the `start` path that emits the outcome after hooks, `finish()` is never called on the progress tracker for the post_start phase (`rust/crates/vibe-core/src/hooks.rs:126` context, `start.rs` never reaches `finish()` on that path), so the phase is left frozen mid-spinner and `closing_outcomes` never runs.

The reviewer of PR #605 explicitly scoped this out of the glyph fix and asked for a follow-up issue. Note the abandoned-task rendering added in #605 (`☐` dim) does NOT resolve this: the defect is that `finish()` is not called at all on that path.

## Suggested fix

Ensure every exit path of `start` (including the outcome-emitting path after hooks) finalizes the tracker, and add a test asserting `finish()` is reached.

Severity: low-medium / reported by code review of PR #605.

Found during automated review (Claude Code).

kexi:
Fixed by #607 (merged after this issue was filed). `run_config_and_hooks` now finalizes the tracker on every non-fatal end, gated path included (`rust/crates/vibe-core/src/commands/start.rs:835-836`), and a failing `post_start` is warn-and-continue rather than an early error return (`start.rs:1002-1019`), so `finish()` is reached on the outcome-emitting path and `closing_outcomes` runs.

The suggested regression tests also landed in #607 and pass on current develop:
- `tracker_finishes_when_a_hook_fails` (`rust/crates/vibe-core/src/commands/start_tests.rs:2266`) — failing `post_start`, asserts every `Started` tracker event has a matching `Finished`.
- `tracker_finishes_when_the_run_is_gated` (`start_tests.rs:2242`) — same guarantee on the `pre_start`-gated path.

Note: FATAL errors (e.g. a directory-copy failure) intentionally leave the tracker unfinished so pending bars are not stamped with the success glyph above the `Error:` line (`start.rs:830-834`, test `tracker_is_not_finished_when_the_error_is_fatal`); that abort rendering was #600's scope, not this issue's. Closing as fixed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #622, opened 2026-08-16
fix: render phase outcome from child outcomes instead of success-stamping

### LINKED ISSUE #614: fix: phase bars are success-stamped by finish() even when a child task failed
## Summary

`IndicatifTracker::finish` success-stamps parent phase bars unconditionally: phase bars are never explicitly completed or failed anywhere in the codebase, so a phase whose child task called `fail_task` still closes with the success glyph.

Pre-existing on develop (e.g. the `copy_files` path warns per-file, calls `fail_task`, returns `()`, and reaches `tracker.finish()` which success-stamps the parent "Copying files" phase over a failed child). Raised by Codex review of PR #607 and confirmed as out of that PR's diff.

## Suggested fix

Track child outcomes per phase and close the phase bar with a failure/partial glyph when any child failed. Companion to #600 (glyph semantics) and #611 (finish() never reached on one path).

Found during the Codex/Claude review convergence loop (PRs #605/#607).

kexi:
## Summary

A phase bar reports no outcome of its own — nothing calls `complete_task` or
`fail_task` on a phase id — so `finish()` decided its glyph. The rule in
`closing_outcomes` only asked whether a child was still *pending*, and a
`fail_task`'d child counts as done, so a phase whose hook exited non-zero
closed with the success glyph `☒` directly above the child's red `✗`. On the
`copy_files` path (which warns per file, calls `fail_task` and returns `()`),
that made a run with failed copies read as fully successful.

`finish()` now folds the children into the phase verdict, in this precedence:

1. a child is still pending → `⊘` Abandoned (the run ended before the phase's
   result was known);
2. otherwise a child failed → `✗` Failed, reason `(failed: N task(s) failed)`;
3. otherwise → `☒` Completed.

Why pending outranks failure: `⊘` already says "we stopped before knowing", and
promoting the phase to `✗` would assert the remaining children failed too. Why
the phase reason counts tasks instead of quoting a child's error: the child line
right below already carries it verbatim, and a phase with several failures has
no single reason to quote. Skipped children (`☐`) are deliberately not failures
— nothing ran, so they leave the verdict to the tasks that did.

The rule is documented in the `progress.rs` module header and in the marker
legend of the README triplet and `configuration/hooks` docs triplet.

Companion to #600 (glyph vocabulary, reused unchanged) and #611.

## Test Plan

- Rewrote `finish_keeps_a_failed_task_and_still_closes_its_phase` (which pinned
  the buggy `☒` phase line) as `finish_fails_a_phase_whose_task_failed`.
- Added tests in `rust/crates/vibe-core/src/progress.rs` for mixed outcomes:
  success + failure, failure count pluralization, failure + pending (abandoned
  wins), failure scoped to its own phase, skipped-not-a-failure, and the red
  coloring of the aggregated phase line.
- All other rendering tests unchanged and green.
- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`,
  `cargo test --workspace` (1121 tests), `pnpm run fmt:check`,
  `pnpm run check:i18n`, `pnpm run test:npm`, `pnpm run check:docs` all pass.

## Checklist

- [x] Tests added/updated

Fixes #614


kexi:
## Review-fix convergence loop — converged in 1 round

Reviewer: **Codex (gpt-5.6-sol, reasoning effort: high)** · Fixes/triage by **Claude**

### Rounds

| Round | Findings | Fixes | Disputes |
| ----- | -------- | ----- | -------- |
| 1     | `[]` (clean) | — | — |

Round 1 returned an empty finding set, which is the loop's convergence condition, so no
further rounds were run. The review was executed twice independently against
`git diff origin/develop...HEAD` and returned `[]` both times.

### Fixes applied

None — nothing to fix. The branch is unchanged from `807d2df`; no new commit was pushed.

### Disputed findings

None raised.

### Independent peer verification

Because a clean round is only as good as its verification, the following properties of
`closing_outcomes` / `close_bar` were checked directly against the code rather than taken
on the reviewer's word:

- `skip_task` sets `done = true` and leaves `failed` at its `push_bar` default of `false`,
  so a skipped child is neither pending nor failed — matching the documented rule that
  skipped tasks leave the phase verdict to the tasks that actually ran.
- `close_bar` reassigns `failed` from the outcome on every close, so the flag cannot go
  stale if a node is closed more than once.
- `NodeKind::Task { phase }` resolves its parent via `tallies.get_mut(phase)`, so an id
  from another tracker or an out-of-range parent matches nothing instead of panicking; the
  second pass only reads a tally inside the `NodeKind::Phase` arm.
- `closing_outcomes` returns `None` for already-closed nodes, so `finish()` never repaints
  over a child's verbatim `✗ … (failed: <error>)` line.
- The synthesized phase reason is program-generated and still passes through
  `render_line`'s `sanitize_for_display`; keeping a `failed: bool` instead of re-parsing the
  rendered message means the phase verdict never depends on attacker-influenced text.
- The precedence documented in README (en/ja/zh) and `configuration/hooks.mdx` (en/ja/zh)
  matches what `closing_outcomes` implements.

### Checks

All green on `fix/614-phase-failure-rendering`:

- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace` — 1121 passed, 0 failed
- `pnpm run check:i18n` — en / ja / zh in sync

### Follow-up items

None. One cosmetic note for the record, not a defect: the docs render the phase reason as
`(failed: N task(s) failed)` as shorthand for both grammatical forms, while the code emits
`1 task failed` / `2 tasks failed`. The pluralization itself is pinned by
`finish_counts_every_failed_task_in_the_phase_reason`.

PR #623, opened 2026-08-16
fix: match word boundaries at any occurrence in jump

### LINKED ISSUE #331: perf: optimize jump command search to single-pass algorithm
## Context

From PR #324 review feedback (non-blocking suggestion).

## Problem

The current `vibe jump` search logic performs multiple passes over the worktree list:

1. Exact match (case-sensitive)
2. Exact match (case-insensitive)
3. Word boundary match (case-sensitive)
4. Word boundary match (case-insensitive)
5. Substring match (case-sensitive)
6. Substring match (case-insensitive)

## Proposal

Optimize with a single pass that categorizes matches into buckets, then selects the highest-priority non-empty bucket.

## Priority

Low — typical repositories have <10 worktrees, making the current multi-pass approach negligible in practice. This becomes relevant only for repositories with hundreds of worktrees.

## Related

- PR #324

### LINKED ISSUE #332: feat: improve word boundary matching to check all occurrences
## Context

From PR #324 review feedback (non-blocking suggestion).

## Problem

The `isWordBoundaryMatch` function uses `indexOf`, which finds only the **first** occurrence of the search term.

For example, with branch `feat/login-page-login`, searching for `login` would only check if the first occurrence (at position 5) is at a word boundary, not the second occurrence (at position 16).

## Proposal

Consider using segment-based matching:

```typescript
branch.split(/[\/-_]/).some(part => part.includes(search))
```

## Priority

Low — branch names rarely have repeated segments, and the substring fallback catches these cases anyway.

## Related

- PR #324

kexi:
## Summary

`is_word_boundary_match` in `rust/crates/vibe-core/src/commands/jump.rs` only
inspected `branch.find(search)` — the **first** occurrence of the query. A branch
whose earliest occurrence sat mid-word was therefore demoted to the substring
tier even when a later occurrence began a segment.

Concretely, for the query `user`:

| branch           | before          | after           |
| ---------------- | --------------- | --------------- |
| `superuser/user` | substring match | word-boundary match |
| `relogin-login`  | substring match | word-boundary match |
| `myuser_user`    | substring match | word-boundary match |

The user-visible effect is ranking: such a branch competed in the substring tier
against genuinely weaker matches and could trigger an unnecessary selection
prompt instead of jumping straight to the obvious candidate. The fix iterates
`match_indices` and accepts the branch if **any** occurrence follows a boundary
character (`/`, `-`, `_`) or starts the string — which is exactly what the docs
already promise ("If the input appears at a word boundary (after `/`, `-`, or
`_`)", `packages/docs/src/content/docs/commands/jump.mdx`), so no doc change is
needed.

The empty-query case is handled explicitly rather than falling into the loop:
an empty needle matches at byte 0 of every string, which preserves the old
`find("") == Some(0)` result.

Fixes #332

---

### Closes #331 — the tiers stay (decision, with measurements)

#331 proposed replacing the six sequential match tiers with a single scored pass
that buckets candidates. I evaluated it against the bar of "byte-identical
selection for every existing test **and** the doc-comment contract, with a
cleaner implementation", and it does not clear it. Recommending closure as
won't-do:

**1. The win is unmeasurable in context.** Each tier is a single O(n) scan and
`n` is the number of worktrees. Worst case (all six tiers run, zero matches),
measured on this machine with an optimized build of the extracted matcher:

| worktrees | matcher, worst case |
| --------- | ------------------- |
| 10        | 3.2 µs              |
| 100       | 27.6 µs             |
| 1000      | 273 µs              |
| 10000     | 2.76 ms             |

For comparison, the single `git worktree list --porcelain` subprocess that
`jump` must run first costs **~14.6 ms** in this repository. At a realistic 10
worktrees the entire six-tier cascade is ~0.02% of the command's runtime —
about 4500x cheaper than the git call it already pays for, and still ~50x
cheaper at 1000 worktrees. The issue itself notes typical repos have <10
worktrees.

**2. The tier structure is a test-pinned UX contract, not incidental code.**
The tier order is documented in the module header, in
`packages/docs/src/content/docs/commands/jump.mdx` (and its `ja`/`zh` mirrors),
and pinned by the `tier_*_only` cascade-isolation tests, each of which
constructs a worktree list matching for the first time at exactly one tier.
Tiers 1/2 additionally differ from 3–7 in ways a uniform scored pass has to
special-case rather than simplify: exact matching runs over the *full* worktree
list while 3–7 run over the scratch-filtered pool, tier 2 prints an extra
`Matched:` line that tier 1 does not, and tier 7 (fuzzy) sorts its bucket by
`f64` score descending while tiers 3–6 preserve `git worktree list` order.

**3. It is not actually cleaner.** Because the tiers are nested supersets
(CS ⊆ CI, word-boundary ⊆ substring), "first non-empty tier" and "highest-priority
non-empty bucket" do produce the same set — but expressing it as buckets replaces
six self-documenting, individually testable filters with per-candidate tier-index
bookkeeping plus the pool/output special cases above. That trades a structure
that reads like the documentation for one that has to be mentally decompiled back
into it, in exchange for microseconds. The maintainability cost outweighs the win.

Closes #331

## Test Plan

- Added unit tests in `rust/crates/vibe-core/src/commands/jump.rs`:
  - `word_boundary_helper_checks_every_occurrence` — a later boundary occurrence
    counts for each of `/`, `-`, `_`; the issue's `feat/login-page-login` repeated
    segment still matches; a branch with **no** boundary occurrence
    (`relogin-xlogin`) is still rejected.
  - `tier_word_boundary_reached_when_only_later_occurrence_is_at_boundary` —
    end-to-end, `superuser/user` now resolves at the word-boundary tier.
  - `word_boundary_tier_outranks_substring_for_later_occurrence` — with a
    competing substring-only branch present, the boundary match wins outright and
    **no** selection prompt is shown.
- `cd rust && cargo fmt --all --check && cargo clippy --workspace --all-targets -- -D warnings && cargo test --workspace` — all pass (1037 vibe-core tests, 0 failures); all pre-existing `tier_*` cascade tests unchanged and green.
- `pnpm run test:npm` (301 passed), `pnpm run fmt:check`, `pnpm run check:i18n` — pass.
- `pnpm run test:e2e` could not run in this sandbox: node-pty fails with
  `posix_spawnp failed` for any command (reproduced with `/bin/echo`), so it is an
  environment limitation, not a regression. This change is Rust-only and the e2e
  suite has no coverage of the `jump` matching tiers; CI will run it.

## Checklist

- [x] Tests added/updated
- [x] `pnpm run check:all` passes (except `test:e2e`, blocked by the sandbox pty limitation noted above)
- [x] Docs updated (if needed) — not needed; the fix makes the code match the existing documented behaviour


kexi:
## Review–fix convergence loop: converged after 2 rounds

Reviewer: **Codex (gpt-5.6-sol, reasoning effort high)**, read-only sandbox, fresh context each round.
Fixes: **Claude**. Each finding was triaged as a peer rather than applied blindly.

| Round | Findings | Outcome |
| ----- | -------- | ------- |
| 1 | 1 (minor) | Verified as a real bug → fixed in `51cfb6c` |
| 2 | 0 | Clean — converged |

### Round 1 finding — accepted and fixed

**`match_indices` enumerates only non-overlapping matches**, so the round-1 implementation still missed boundary occurrences in exactly the class of case this PR set out to fix.

`str::match_indices` resumes each search *past the end* of the previous hit. When an earlier mid-word occurrence overlaps a later boundary one, the boundary occurrence is never yielded. Verified empirically before accepting:

```
"xa/a/a" / "a/a": match_indices=[1]     all occurrences=[1, 3]
"aaa"    / "aa" : match_indices=[0]     all occurrences=[0, 1]
```

Index 1 is mid-word and rejected; the match at index 3 follows `/` and should have qualified, but was never seen — so `xa/a/a` was still demoted to the substring tier. This affects both call sites (the case-sensitive tier and the case-insensitive tier, which share the helper).

The fix scans every char boundary and tests `starts_with`, which sees overlapping and non-overlapping occurrences alike. Iterating `char_indices()` rather than raw byte offsets also keeps the `branch[..index]` slice on a char boundary, so multi-byte branch names cannot panic.

Regression tests added:

- `word_boundary_helper_finds_overlapping_occurrence` — the `xa/a/a` case plus the `-` / `_` variants, and `("xaaa", "aa")` still correctly rejected.
- `word_boundary_helper_handles_multibyte_branches` — `café/feature`, `日本語/ブランチ` and their negative counterparts, covering the slice-panic risk.

All expected values were verified against a standalone harness before being written as assertions, including the pre-existing cases to confirm no behavioural regression.

### Disputed findings

None — the single finding raised was legitimate.

### Verification

`cd rust && cargo fmt --all --check && cargo clippy --workspace --all-targets -- -D warnings && cargo test --workspace` — all pass. **1039** vibe-core tests, 0 failures (1037 before, +2 new); every pre-existing `tier_*` cascade test unchanged and green.

### Follow-ups

None outstanding. No pre-existing or out-of-diff issues were surfaced.

PR #624, opened 2026-08-16
fix: refuse stale release runs and pin the npm dist-tag explicitly

### LINKED ISSUE #616: fix: re-running an old unpublished release run (or a publish-npm re-fire) can roll distribution channels backwards
## Summary

Two pre-existing rollback hazards in the release pipeline, deliberately left out of PR #608's scope (its tap-rollback guard covers the `should_release=false` resume path only):

1. **Old unpublished run republishes with `--latest`**: re-running an old release run that never published goes through `should_release=true` untouched by #608's guard — it would publish that old version with `--latest` and regenerate `Formula/vibe.rb` backwards.
2. **publish-npm.yml `workflow_run` re-fire**: a re-fired publish run can move the npm `latest` dist-tag back to an older version. `publish-npm.yml` is untouched by #608.

## Suggested fix

Extend the currency guard to the `should_release=true` path (compare the version being released against the newest existing release before `--latest`/tap/npm-tag operations), and gate the npm dist-tag step on the version being `>=` the current `latest`.

Constraints: OIDC/GITHUB_TOKEN only, SHA-pinned actions, gate logic in testable scripts.

Raised by Codex review of PR #608 (rounds 3-4), verified pre-existing on develop.

Found during the Codex/Claude review convergence loop.

kexi:
## Summary

Fixes #616

Two rollback hazards left out of PR #608's scope (its tap-rollback guard covers the `should_release=false` resume path only).

### Hazard 1 — an old unpublished run republishes with `--latest`

A release run that built its artifacts but never published can be re-run at any later date. `prepare` sees no published release for its tag, reports `should_release=true`, and the run takes the ordinary path — including `gh release edit --draft=false --latest`, which moves GitHub's latest pointer back to that old version. #608's tap guard reads `isLatest` **after** that edit, so it sees the stale release as current and mirrors it into `Formula/vibe.rb`. The guard is not wrong; it is asking the right question at the wrong time.

- New `scripts/assert-release-currency.ts` — fail-closed gate. Reads `gh release list --json tagName` on stdin (offline, testable, and the API call stays where its auth already lives) and refuses a version older than the newest existing release. Equal is allowed, so #598's idempotent resume path is untouched.
- Wired into `release.yml`'s `prepare` on the `should_release=true` path — before anything is built — **and** re-asserted in the `release` job immediately before `gh release edit --draft=false --latest`. Hours of builds sit between the two, and at the second point the draft still holds no tag, so failing there publishes nothing and the next attempt reclaims the draft.
- **Fail closed, not "publish without `--latest`"**: the repo has no documented back-release use case (`release.yml` refuses off-main dispatches; a tag existing without a release is already fatal), so an older version reaching that point means a stale run. Publishing it half-configured would burn its tag for real and leave a release nobody asked for; refusing costs one re-dispatch.
- Mirrored in `beta-release.yml`'s `prepare`, comparing against **prereleases only** — a stable `3.1.0` outranks `3.1.0-beta.N` by semver and develop keeps its base version until the next bump, so comparing against stables would refuse every ordinary beta cut made after a release. Its tap guard (#608) would catch a stale beta downstream, but only after the prerelease is public and its tag burned.

### Hazard 2 — `publish-npm.yml` `workflow_run` re-fire

`npm publish` with no `--tag` writes the `latest` dist-tag, and a completed `workflow_run` can be re-fired long after the release it followed. Nothing in the publish path consulted the registry.

- New `scripts/resolve-npm-dist-tag.ts` — resolves `latest` only when the version is `>=` the registry's current `latest`, otherwise `previous`. Both publish steps (the five platform packages and the `@kexi/vibe` shim) now pass `--tag` explicitly.
- **Non-default tag rather than failing**, unlike hazard 1: republishing an old version to npm is a legitimate repair — a platform package whose first publish failed still needs its bytes on the registry, just not on `latest`. `previous` keeps them installable by exact version.
- A package with no `latest` yet (`npm view` prints nothing) claims `latest`; anything unorderable on either side fails closed onto the non-default tag.
- OIDC-only auth is unchanged — no `NODE_AUTH_TOKEN`, no new secrets.

### Shared ordering

Both gates and the beta mirror sit on one comparator (`scripts/release-version-order.ts`) so they cannot drift apart. It parses only the two shapes this project publishes (`X.Y.Z`, `X.Y.Z-beta.N`), compares components numerically (`3.10.0 > 3.9.0`, which a lexical sort gets backwards), sorts a `-beta.N` below its own stable, and orders `3.2.0-beta.1 > 3.1.0-beta.99` — the case a bare ordinal compare gets wrong across a base-version bump. Why not a semver library: it would accept build metadata and arbitrary prerelease identifiers no gate here should reason about, and pulling a dependency into the release path widens its supply chain for a numeric compare.

No third-party action `uses:` lines were added or changed. `beta-release.yml`'s `prepare` gains `./.github/actions/setup-toolchain` (a local composite) because the gate needs `bun`.

## Test Plan

- New `packages/npm/test/release-version-order.test.ts` (21 tests): the two accepted shapes and the rejected ones (`-rc.1`, `+build.5`, trailing newline, `v` prefix), numeric component ordering, beta-before-its-stable, `3.2.0-beta.1 > 3.1.0-beta.99`, and `newestVersion` returning the maximum irrespective of input order — `gh release list` sorts by creation time, and a re-run that recreates a release gets a fresh timestamp.
- New `packages/npm/test/assert-release-currency.test.ts` (21 tests): the #616 regression itself (3.1.0 refused while 3.2.0 exists), the current version still re-releasable (#598), the first release, `3.9.0` refused after `3.10.0`, the beta ordinal cases, and — the load-bearing ones — empty stdin / malformed JSON / a non-array / a missing `tagName` all being hard errors rather than reading as "no releases exist", which is exactly the reading that lets a stale run through.
- New `packages/npm/test/resolve-npm-dist-tag.test.ts` (13 tests): older-than-latest resolves to `previous`, newer and equal resolve to `latest`, no-registry-latest resolves to `latest`, both unorderable sides fail closed, and the fallback tag stays a plain identifier npm will not read as a semver range.
- `pnpm run test:npm` — 356 tests pass across 16 files.
- Both CLIs smoke-tested end to end under `set -euo pipefail` in the exact `RELEASE_LIST=$(...)` / `printf | bun` shape the workflows use, including that a refusal propagates out of the pipeline as exit 1.
- `actionlint` on all three workflows: the finding set is byte-identical to the pre-change baseline (4 pre-existing shellcheck info/style notes on lines this PR does not touch), so nothing new was introduced.
- `pinact run --check`, `pnpm run fmt:check`, `pnpm run lint`, `pnpm run check:i18n` pass. No Rust files changed.

There is no runnable harness for a workflow job graph, so the enforcement is the extracted, tested gate scripts; full CI validates this PR.

## Checklist

- [x] Tests added/updated
- [x] `pnpm run check:all` passes (targeted checks run locally; full CI validates on this PR)
- [x] Docs updated — `CONTRIBUTING.md` and the release skill now state that a never-published run cannot be resurrected later, and that a repair publish lands on `previous`


kexi:
## Review convergence loop — 3 rounds

Reviewer: **Codex (gpt-5.6-sol, reasoning effort high)**, read-only, fresh context each round. Fixes by **Claude**.

Three findings were raised; **all three were accepted and fixed**. Both fixes landed in the guards' own fail-safe logic — the review found the gate itself had fail-open paths.

### Round 1 — 2 findings, both fixed (`a6295be`)

**1. `npm view … || true` was fail-open (major).** The dist-tag gate collapsed every `npm view` failure into an empty string, which the resolver reads as "no `latest` yet, so this publish creates it". But an absent package and an unreachable registry *both* exit 1 with empty stdout — verified locally:

```
npm view @kexi/nonexistent dist-tags.latest  -> exit 1, empty stdout
npm view @kexi/vibe --registry=http://127.0.0.1:9 -> exit 1, empty stdout
```

So a transient registry outage during a re-fired `workflow_run` would have handed `latest` to a stale version — precisely the rollback #616 exists to prevent, in the one place that must fail closed.

Fixed by reading `npm view --json` and keeping its exit code. `--json` yields a machine-readable `error.code`, so only a confirmed `E404` means "no latest"; anything else aborts the step. The interpretation lives in a new exported `interpretNpmView()` rather than in shell, so it is unit-testable — the previous shape had **no test covering exit-code handling at all**. Verified end-to-end: E404 → `latest`, ECONNREFUSED → step fails with `::error::… refusing to treat a registry failure as an absent dist-tag`.

**2. No `concurrency` on `publish-npm.yml` (major).** The gate reads the registry in one step and publishes in the next, with nothing serializing runs (`release.yml` and `beta-release.yml` both already had groups). Two overlapping runs could read the same pre-publish `latest` and both claim it, the older winning if it finished second. Added `concurrency: {group: publish-npm, cancel-in-progress: false}`.

### Round 2 — 1 finding, disputed → **overturned in round 3**

Codex argued the new group does not fully serialize, since a concurrency group keeps only one pending run, and proposed `queue: max`. I disputed this, primarily because `actionlint` 1.7.12 rejects `queue` as an unknown key.

**I was wrong, and the round-3 rebuttal was correct.** `queue: max` is genuine GA GitHub Actions syntax [added 2026-05-07](https://github.blog/changelog/2026-05-07-github-actions-concurrency-groups-now-allow-larger-queues/); [the docs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency) confirm `queue` defaults to `single` (one pending run, later arrivals **evict** it), `max` allows 100 in FIFO, and it is valid only with `cancel-in-progress` false/unset. actionlint's latest release (v1.7.12, March 2026) simply predates the feature — a lint lag, not invalid syntax.

### Round 3 — disputed finding re-evaluated and **fixed** (`9c04386`)

The eviction is reachable here, and it silently drops a real publish. This workflow triggers on `workflow_run: completed`, which fires for **failed** Release runs too. Those skip every job (`resolve-release` is gated on `conclusion == 'success'`, everything else `needs:` it) — but a skipped run still enters the group. So *running + pending + a late no-op event* evicts the pending publish, leaving a published GitHub Release with **no npm version behind it**. Only three runs in the group are needed, not three concurrent publishers.

Added `queue: max`, with a comment recording the actionlint caveat.

### Verification

- `pnpm run test:npm` — **362 passed**, including new `interpretNpmView` cases for the E404 / ECONNREFUSED / malformed-body split.
- `actionlint` — compared against an `origin/develop` baseline normalized for line-number drift: the 5 pre-existing findings are unchanged, and the **only** new one is the expected `unexpected key "queue"` false positive. actionlint is not wired into this repo's CI.
- `pinact run --check` — clean.

### Convergence

Converged on substance: no finding remains open or disputed. Rounds 1 and 3 each closed a genuine fail-open path; round 2's finding was upheld on appeal and fixed. The 3-round cap was reached, so no clean no-findings round was run — worth noting, though the only open item is external.

### Follow-up (not blocking)

Once an actionlint release supports `concurrency.queue`, the caveat comment in `publish-npm.yml` can be dropped. If actionlint is ever added to CI, it must be a `queue`-aware version or this workflow will fail lint on valid syntax.

PR #625, opened 2026-08-16
chore: resolve the six audit follow-up findings from #603

### LINKED ISSUE #603: chore: low-priority findings and coverage gaps from parallel code audit
Collected low-severity / uncertain findings from the 2026-08-16 parallel code audit (Claude Code). None are release-blocking; tracked here so they aren't lost. Related confirmed bugs: #596 #597 #598 #599 #600 #601 #602.

## Low severity (confirmed)

- **`Formula/vibe.rb:30-35`, `Formula/vibe-beta.rb:31-34`** — the 4 modifier-`if` assignments leave `binary_name` as `nil` if none match, so `bin.install nil => "vibe"` raises `TypeError`. `vibe-versioned.rb:31-41` handles the same logic with `if/elsif/else + odie` — the formulas are asymmetric. Low reachability (the `on_macos`/`on_linux` url blocks constrain earlier).
- **`.github/workflows/publish-npm.yml:24`** — `id-token: write` is declared at workflow level, granting OIDC issuance to non-publish jobs (`verify-versions`, `security-audit`) that run `cargo install` / `pnpm install`. Scope it down to the two publish jobs.
- **`package.json` `check:docs`** — runs lint/format/check but not `build`, so a docs build breakage is only caught by `docs.yml` on push to main. Coverage gap, not a bug.

## Uncertain (no concrete failure scenario constructed; needs investigation or tests)

- **`rust/crates/vibe-core/src/commands/clean.rs:243`** — `read_to_string(&git_file_path).ok()` silently skips fast-remove and falls back to the traditional path with no verbose log; no test covers the fast-remove fallback. Test/observability gap.
- **Dual macOS detection** — `detector.rs:35` uses `cfg!(target_os = "macos")` while `strategies.rs:229` `is_macos()` uses `native.get_platform() == "darwin"`. They currently degrade safely together (probe uses the same `cfg!`), but the duplication invites divergence. Unify on one source.
- **Partial-copy debris + fallback merge semantics** — `strategies.rs:294` falls back from a soft error to `standard_copy_directory`, which merges over any debris left by a failed `cp -r` (no cleanup), and `cp -r src dest` nests into `dest/src/` when dest exists while `rsync -a -- src/ dest` copies contents — so the resulting tree can differ by strategy. Worth a cleanup-before-fallback or documented invariant.

kexi:
## Summary

Clears the six verified-still-present findings from the 2026-08-16 parallel code audit umbrella. Each item is its own commit; none change user-visible CLI behaviour except where noted.

- [x] **(a) `Formula/vibe.rb` + `Formula/vibe-beta.rb`** — the four modifier-`if` assignments left `binary_name` as `nil` when no branch matched, so `bin.install nil => "vibe"` raised an opaque `TypeError`. Both now use the `if/elsif/else` + `odie "Unsupported platform"` shape `vibe-versioned.rb` already had, making the three formulas symmetric.
- [x] **(b) `.github/workflows/publish-npm.yml`** — `id-token: write` moved from workflow level to job-level `permissions` on `publish-platform-packages` and `publish-cli` only. `verify-versions` and `security-audit` run third-party code (`pnpm install`, `cargo install`) and never issue an OIDC token; they now get an explicit `contents: read`. No new secrets, no action changes.
- [x] **(c) `package.json` `check:docs`** — now delegates to the docs package's existing `ci` script (`lint && format && check && build`), so a docs build breakage is caught locally / in `check:all` instead of only by `docs.yml` after the push to main. Reusing `ci` keeps the docs gate defined in one place. Cost: ~3 min (88 pages).
- [x] **(d) `rust/crates/vibe-core/src/commands/clean.rs`** — `read_to_string(.git).ok()` discarded the error, so a clean that silently fell back to the slower traditional path gave no clue why even under `--verbose`. The fallback itself is correct (without the link file content, fast remove cannot restore the worktree after moving it); this only names the cause, and adds the test that branch never had.
- [x] **(e) Dual macOS detection** — `detector.rs` decided `cp -c` vs `cp --reflink=auto` with `cfg!(target_os = "macos")` while `strategies.rs::is_macos()` decided the same thing from `NativeClone::get_platform()`. Unified on `copy::native::is_darwin`, sourced from `vibe-native`'s `get_platform()`. Chosen over `cfg!` because that side cannot be driven by `FakeNative`, so a divergence between the probed capability and the argv the strategy actually runs would have been untestable.
- [x] **(f) `copy/strategies.rs` partial-copy debris** — a soft failure from `cp -r` / `rsync` / `clonefile` can leave a half-written tree at `dest`, and `standard_copy_directory` copies *over* an existing tree, so the fallback silently merged stale files that never existed in `src` into the new worktree. The debris is now removed before falling back (best-effort; skipped when `dest` predates the copy, so pre-existing user content is never wiped), and the abandoned strategy's error is folded into the fallback's so the root cause still reaches the caller's warning.

The `UnsupportedFileType` hard-error rule (finding #5, CWE-59) is untouched — it still returns before any cleanup or fallback, and `native_symlink_rejection_does_not_fall_back` still asserts `dest` is never created.

## Test Plan

Added:

- `commands/clean_tests.rs::unreadable_git_file_falls_back_to_traditional_remove_with_verbose_reason` — guarantees the degraded path still removes the worktree and that `--verbose` names the unreadable `.git`.
- `copy/native.rs::platform_tests` — the macOS predicate matches only `"darwin"`, and the probe's platform source agrees with `RealNativeClone`.
- `copy/strategies.rs` — debris is discarded before the fallback; a pre-existing destination survives; a failing fallback reports the original strategy error too.

Verified `soft_failure_discards_partial_copy_debris_before_falling_back` fails on the un-fixed code (`debris from the failed strategy must not survive into the fallback copy`), so it is not vacuous.

Ran:

- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (1121 tests, all pass)
- `pnpm run test:npm` (301 pass), `fmt:check`, `lint`, `check:i18n`, `check:licenses`
- `pnpm run check:docs` including the new build (88 pages, exit 0)
- `actionlint .github/workflows/publish-npm.yml` (only the pre-existing SC2129 style note, identical on `develop`) and `pinact run --check` (clean)
- `ruby -c` on all three formulas

`pnpm run test:e2e` could not be validated here: every node-pty spawn fails with `posix_spawnp failed` in this sandbox. Confirmed identical on unmodified `origin/develop` (71 failed there too), so it is an environment limitation, not a regression — CI will cover it.

## Checklist

- [x] Tests added/updated
- [x] `pnpm run check:all` passes (except `test:e2e`, blocked by the sandbox as described above)
- [x] Docs updated (if needed) — no user-facing docs change

Fixes #603

## Summary by CodeRabbit

* **Bug Fixes**
  * Installers now provide a clear error on unsupported operating systems or CPU architectures.
  * Improved compatibility when selecting platform-specific copy and clone methods.
  * Failed directory copies now clean up eligible leftover files while protecting existing destinations and symlinks.
  * Cleaning operations provide more useful verbose diagnostics when worktree metadata cannot be read.

* **Reliability**
  * Copy and cleanup operations now preserve the original error when fallback recovery also fails.


kexi:
## Review–fix convergence loop: 3 rounds, converged

Reviewer: **Codex (gpt-5.6-sol, reasoning effort high)** · Fixes and triage: **Claude**

Three review rounds were run against `git diff origin/develop...HEAD`, each a fresh `codex exec` carrying its own context (no resume), with disputed findings fed back for explicit re-evaluation.

### Findings fixed (5)

| # | Round | Sev | Finding | Fix |
|---|-------|-----|---------|-----|
| 1 | 1 | major | `dest_preexisting` used `Path::exists()`, which dereferences — a pre-existing **dangling symlink** at `dest` read as absent, so the new debris cleanup would unlink it | `std::fs::symlink_metadata(dest).is_ok()` (`ec72a09`) |
| 2 | 2 | major | `remove_dir_all` does not follow a symlink in the *final* component, but path resolution still traverses **intermediate** ones — a racing worker on nested `copy.dirs` could make `dest`'s parent a link and the cleanup would delete an unrelated real tree | extracted `discard_own_debris()`, which declines when `dest` is a link/non-directory **or** its parent is a link (`8c2426c`) |
| 3 | 2 | minor | the platform tests were **tautological**: both sides of `host_platform_matches_the_native_seam` call `vibe_native::get_platform()`, so a rename of `"darwin"` would keep them equal while `is_darwin(host_platform())` silently went false and flipped macOS to `cp --reflink=auto` | added `is_darwin_agrees_with_the_host_target_os`, anchored to `cfg!(target_os = "macos")` (`8c2426c`) |
| 4 | 3 | minor | `select_directory_strategy` still compared `platform == "darwin"` directly, so the "single decision point" contract in `is_darwin`'s own doc comment did not hold | routed through `is_darwin` (`63957e3`) |
| 5 | 3 | major (partial) | the parent-only guard misses a link **further up** the prefix (`a/link/mid/subdir`) | see "Known limit" below (`63957e3`) |

Each regression test was verified to **fail before** its fix and pass after.

### Known limit, deliberately pinned rather than closed

Finding 5 is real but cannot be fixed proportionately here. Catching a link above the immediate parent needs a boundary to stop an ancestor walk at — the worktree root — which `copy_directory` is not given. Every self-contained stopping rule was tried and fails on macOS: `/var` and `/tmp` are themselves symlinks, and **no ancestor of a temp path is its own canonical form**, so a root-ward walk refuses *every* cleanup (it broke the honest-path tests). Requiring `canonicalize(dest) == dest` fails for the same reason.

So the gap is now **stated in `discard_own_debris`'s doc comment and pinned by `debris_cleanup_still_follows_a_symlink_deeper_in_the_prefix`**, which asserts the *current* behaviour — the follow-up that plumbs the worktree root through the executor will see that test flip. Residual exposure needs nested `copy.dirs` more than one level apart, a soft strategy failure, *and* the link to appear inside the check→remove window.

### Findings disputed (2, both twice)

- **Staging directory + atomic rename instead of in-place cleanup.** A design rewrite beyond this PR's scope. The fallback deliberately copies *over* an existing tree (pinned by `fallback_preserves_a_destination_that_predates_the_copy`), and a directory rename is not an atomic replace over a populated destination anyway. The one concrete data-loss mechanism raised under this heading (intermediate symlink) was fixed directly instead.
- **`let _ = remove_dir_all(...)` swallows the cleanup error.** `develop` had *no* cleanup at all and always merged over debris, so every case that succeeds today still succeeds; making cleanup failure fatal would **add** a failure mode and regress against `develop`. The swallowing is explicit and documented. Codex did not re-raise this after round 1.

An empirical check also retired the original escalation here: `remove_dir_all` on a symlink-to-a-directory removes only the link and leaves the target's contents intact, and on a plain file returns `NotADirectory` without deleting anything.

### Follow-up items

1. Plumb the worktree root into `RealCopyExecutor` so `discard_own_debris` can bound an ancestor walk and cover the whole prefix (flips the pinned test).
2. Pre-existing, untouched by this PR: `publish-npm.yml` SC2129, `beta-release.yml` / `release.yml` SC2086, `nix.yml` unused `nix-paths` anchor.

### Checks

`cargo fmt --check` · `cargo clippy --workspace --all-targets -D warnings` · `cargo test --workspace` (**1045 passed**) · `pnpm run test:npm` (301 passed) · `ruby -c` on both formulas · `actionlint` (5 findings, all pre-existing on `develop` — no new ones) · `pinact run --check` — all green.

**Convergence:** reached at the round cap, with the only open item a documented, test-pinned limit plus its follow-up.

PR #630, opened 2026-09-02
docs: document --allow-default-branch on clean and rename

kexi:
## Summary

PR #584 shipped the default-branch guard for `vibe clean --delete-branch` and `vibe rename` (#578), but the escape hatch `--allow-default-branch` was only mentioned in the changelog; the `clean` and `rename` command pages never explained it. This PR fills that gap, in all three locales (en / ja / zh):

- **`commands/clean.mdx`** — adds the `--allow-default-branch` option and a "Default branch protection" section: the soft-skip behaviour (worktree removed, branch kept, exit 0), the detection order (`refs/remotes/origin/HEAD` → `init.defaultBranch` → `master`), and the caveat that `origin/HEAD` is only written by `git clone`, so a default branch changed later on the remote is unguarded until `git remote set-head origin -a` is run.
- **`commands/rename.mdx`** — adds the option to the table, a "The default branch is protected" section with the error output, and lists the default branch in the exit-code conditions.
- **`vibe-core/src/git.rs`** — drops the stale claim in `resolve_default_branch`'s doc comment that `start` also uses the guard (only `clean` and `rename` do).

Body headings deliberately carry no `<Badge>`: Starlight appends a trailing `-` to the slug of a heading that contains one, which would have broken the in-page and cross-page anchors. The badge stays on the option heading, which nothing links to.

Related to #578.

## Test Plan

- `pnpm run check:all` passes (fmt, lint, i18n, rust fmt/clippy/tests, licenses, npm, e2e, docs build).
- Verified in the built `dist/` that the new heading ids (`default-branch-protection`, `デフォルトブランチの保護`, `默认分支保护`) match the anchors used from the option section and from the `rename` pages.
- Before writing the docs, exercised the real debug binary against scratch repositories to confirm the documented behaviour: soft-skip on a cloned repo, `--allow-default-branch` deleting the branch, `rename` refusing with the shown message, and the stale-`origin/HEAD` case leaving the real default branch unguarded until `git remote set-head origin -a`.

## Checklist

- [ ] Tests added/updated (docs-only; existing unit + e2e coverage from #584 unchanged)
- [x] `pnpm run check:all` passes
- [x] Docs updated (if needed)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GGinK4fsB4xi9M9UJLvyJT

## Summary by CodeRabbit

- **Documentation**
  - Updated `clean` command documentation in English, Japanese, and Chinese to explain default branch protection.
[...]

PR #631, opened 2026-09-02
revert: remove the default-branch guard, and record the v4.0.0 changelog

kexi:
## Summary

Removes the default-branch protection added in #584 (issue #578, shipped in v3.1.0), along with the `--allow-default-branch` flag on `vibe clean` and `vibe rename`, and brings the changelog up to date for the v4.0.0 release.

**Why the guard goes.** It inferred the default branch from `refs/remotes/origin/HEAD` — a ref `git clone` writes and `git fetch` never updates afterwards; refreshing it takes an explicit `git remote set-head origin --auto`, which nobody has a reason to run. Measured across the 109 repositories under `~/ghq/github.com`, comparing each clone's local `origin/HEAD` against `git ls-remote --symref origin HEAD`:

| Outcome | Count |
| --- | ---: |
| local `origin/HEAD` matches the remote's real HEAD | 97 |
| local `origin/HEAD` disagrees | 7 |
| remote unreachable (local value used as-is) | 5 |
| no local `origin/HEAD` at all | 0 |

Every one of the 7 disagreements was the same shape — a repository whose real default had moved to `develop` while `origin/HEAD` still said `main`:

```
CureApp/nash2                        local=main    remote=release/4.0.29
CureApp/sc                           local=master  remote=develop
kexi/ai-travel-planner               local=main    remote=develop
kexi/buta-code-test                  local=buta-…  remote=main
kexi/expo-monorepo                   local=main    remote=develop
kexi/macos-programming-font-preview  local=main    remote=develop
kexi/whopay                          local=main    remote=develop
```

So the guard failed systematically on exactly the `develop`-based workflow it mattered most for, while its name and its message told the user they were protected. It also protected `main` in those repositories — there, not the branch worth protecting. Both "it stopped me" and "it did not stop me" are correct behavior, and neither is explainable without the paragraph above: a permanent support burden bought with a false promise.

**What still protects a shared branch.** The remaining guards answer from local state git fully controls rather than from inference:

- `git branch -d` refuses an unmerged branch, and git refuses to delete a branch checked out in another worktree. `clean` has always relied on this.
- `vibe rename` refuses a branch with a remote upstream. Every shared default branch is pushed, so this already covers the case the removed guard was aiming at — verified against the real binary, where `rename` on a pushed `develop` is now refused as pushed rather than as default.

**What is kept.** `resolve_default_branch` survives with `vibe list` as its only caller, where the same staleness costs a wrong BASE label and nothing more. Its docs now state that narrow contract and say not to build a decision — anything that refuses, deletes or rewrites — on top of it again. `get_default_branch`, the guard-only wrapper, is deleted.

## Changelog for v4.0.0

The changelog jumped from v3.1.0 straight to this removal, leaving the ~90 commits already on develop undocumented. This PR adds the v4.0.0 section covering all of it:

- **Added** — `vibe list`'s BASE/AGE/STATUS columns, its filter and sort flags, the opt-in `[summary]` section, and hook mode's gated-`pre_start` stderr signal.
- **Changed (breaking)** — the config-semantics revision. Making `*_prepend` / `*_append` effective where they were previously parsed and ignored turns dormant config into config that executes, and the trust hash covers bytes rather than interpretation, so affected `.vibe.toml` files now require re-running `vibe trust`.
- **Removed (breaking)** — this PR's guard and flag.

Two independent breaking changes, hence v4.0.0 rather than v3.2.0.

## Test Plan

- `pnpm run check:all` passes: 1056 vibe-core unit tests, the binary suites, 388 npm tests, 65 e2e (1 skipped), docs build with 0 errors, i18n in sync.
- Guard tests are replaced by removal-regression tests rather than deleted: `clean` deletes a branch `origin/HEAD` calls the default and emits no default-branch message; `clean` no longer issues `symbolic-ref` at all; `rename` renames an unpushed default-looking branch; `rename` still refuses a pushed one, now via the pushed-branch guard.
- `packages/e2e/default-branch.test.ts` deleted.
- Re-ran the scratch-repository scenarios against the rebuilt binary: `clean --delete-branch` now deletes `main` and `develop` in a clone, `--allow-default-branch` errors as an unknown argument, and `rename` on a pushed `develop` prints the pushed-branch error with the manual git commands.

## Checklist

- [x] Tests added/updated

Reverts #578. Supersedes #630, which documented the guard and was closed.

https://claude.ai/code/session_01GGinK4fsB4xi9M9UJLvyJT


packages/docs/src/content/docs/ja/changelog.mdx
```diff
@@ -5,6 +5,12 @@ description: vibeのバージョン履歴とリリースノート
 
 vibeの主要な変更はここに記録されています。
 
+## 未リリース
+
+### 削除
+
+- **破壊的変更:** v3.1.0 で追加した既定ブランチの保護を削除しました。`vibe clean` と `vibe rename` の `--allow-default-branch` も削除され、指定するとエラーになります。この保護は既定ブランチを `refs/remotes/origin/HEAD` から推測していましたが、この ref は `git clone` のときに書き込まれたきり更新されません。ローカルの 109 リポジトリで計測したところ 7 件でリモートの実際の既定ブランチと食い違い、そのすべてが「実際は `develop` に移っているのに `origin/HEAD` は `main` のまま」という同じ形でした。つまり最も効いてほしい運用でこそ外れながら、保護されていると表示していたことになります。未マージのブランチや他の worktree でチェックアウト中のブランチは引き続き `git branch -d` が削除を拒否し、push 済みのブランチは引き続き `vibe rename` が拒否します。共有されている既定ブランチはすべてこれで守られます。
```
kexi: Fixed in 7a91b3d. The claim was wrong as written: `git remote set-head origin --auto` does refresh `refs/remotes/origin/HEAD`; what never happens is an automatic update during `git fetch`. Reworded all five sites plus `git.rs`'s `resolve_default_branch` docs, which carried the same claim and was not in the list.

This does not change the argument for removing the guard — an inference that silently goes stale until someone runs a command they have no reason to run is still the wrong basis for refusing an operation — but a reader who wants to correct their own repository can now see which command does it.

